PR #90, opened 2021-12-18
Moonpile/issue89

Moonpile:
- Made "Improve" context menu on skills always appear
- Dialog with up to four options for improvement


wake42 (changes requested):
Hi!
I have some comments. First I don't think we should give page references to the core book since the user might not have that one - they might be playing with the starter-set or even the quick start rules.

Second I think it should be radioboxes instead of checkboxes since checkboxes imply that you can select many while radioboxes imply that only one is selectable.

Finally I think the chance calculation is wrong. If I understand correctly you add the skill category bonus to the chance. In the rulebook they add that bonus to the roll instead. I was thinking about that though... What do you think about changing it so that you have to roll under 100 - learned chance instead (learned chance is the same thing as chance - skill category bonus on the skill item). It seems more logical to succeed a roll to improve instead of failing it, even if it's not exactly according to how the rulebook puts it.

Moonpile:
I think the page reference was removed before I even made the Pull Request.  Perhaps it's in one of the incremental commits?

The experience roll calculation is 1d100+ categoryMod, and if this is greater than the skill chance (total) or greater than or equal to 100, then you gain.

A few things I think we may want to think through better in the future:

- if you add "content" to a roll it "overwrites" the actual dice result.  Therefore on rolls if we have more to say than just a little bit we need to put it in the "flavor" with some HTML around it to make it look nice.  We should come up with a standard for this, or perhaps even an HBS template.
- I put some "style" elements on the divs in the dialog.  If we do more dialogs perhaps we can extract our look and feel to scss.

wake42 (commented):
I think it's more interesting to show the chance of success than the actual mechanics behind it. What do you think about this sketch with less text. Also with radio buttons instead of checkboxes.
image

wake42:
> * if you add "content" to a roll it "overwrites" the actual dice result.  Therefore on rolls if we have more to say than just a little bit we need to put it in the "flavor" with some HTML around it to make it look nice.  We should come up with a standard for this, or perhaps even an HBS template.

Yes the styling and general handling of chat messages as a whole would benefit from a work trough!

Moonpile:
I had to put the chance to gain on another line or the experience h2 element wrapped and the training one didn't and that looked bad.

image
image
image
image

src/items/skill-item/skillSheet.ts
```diff
@@ -108,4 +111,179 @@ export class SkillSheet extends RqgItemSheet<ItemSheet.Options, SkillSheetData |
...
+    var chanceToGain = 0;
+    if (chance > 100) {
+      chanceToGain = categoryMod;
+    } else {
+      chanceToGain = (100 - chance) + categoryMod;
+    }
+
+    const btnImprove = getGame().i18n.format("DIALOG.improveSkillDialog.btnDoImprovement");
+    const btnCancel = getGame().i18n.format("DIALOG.improveSkillDialog.btnCancel");
+    var buttons = {};
```
wake42, line 140: Just some syntax, but it you type this as `const buttons: any = {}` then you don't need the @ts-ignores further down.

Also generally, please don't use `var`. Use `const` wherever possible and `let` if that doesn't compile to keep variable scopes to a minimum :-)

src/items/skill-item/skillSheet.ts
```diff
@@ -108,4 +111,179 @@ export class SkillSheet extends RqgItemSheet<ItemSheet.Options, SkillSheetData |
...
+          await SkillSheet.submitImproveSkillDialog(
+            html as JQuery,
+            actor,
+            item,
+            speakerName
+          )
+      }
+    }
+
+    //TODO: Not sure why this wouldn't work with "await"
```
wake42, line 164: The reason is that the `showImproveSkillDialog` function is not declared as async. Just add that and you can use await.

wake42 (changes requested):
Looking good!
I found that the order of cancel / improve buttons was flipped. Foundry standard is having cancel to the right.
Also I think there is a UX problem if you don't select any radiobutton. Then it just clears the experience without any feedback. I think either disable the submit button until something is chosen, or preselect one option to begin with.
And finally the first "What type of improvement..." line is too long for the row. I would either reformulate to make it shorter, or make the dialog wider.

src/items/skill-item/skillSheet.ts
Moonpile: The limits of my javascript and typescript knowledge are showing!

Moonpile:
Ok, I think I've addressed everything including applying Prettier.

wake42 (approved):
Looking good!

PR #92, opened 2021-12-19
Max HP & MP after character creation

### LINKED ISSUE #87: Max HP & MP after character creation
When an actor is created, or after characteristics are regenereated, the HP and MP of the actor should be max. Today it starts as 0 snd you manually have to raise it to the max value.

Moonpile:
Fixes #87

- When initializing either all characteristics or just one, set the current HP and MP to the max HP and MP
- Manually changing Characteristic values does NOT set current HP or MP to max.
- Added to the warning dialog


wake42 (approved):
Just works - super!

PR #109, opened 2021-12-22
bug combat rows don't align Fixes #108

### LINKED ISSUE #108: combat rows don't align
The line is not aligned
image

Moonpile:
div needed style="display: block"

### DIFF
--- a/src/actors/sheet-parts/gear-tab.hbs
+++ b/src/actors/sheet-parts/gear-tab.hbs
@@ -160,41 +160,41 @@
             <div data-item-id="{{id}}" class="gear contextmenu">{{name}}
               {{#if (eq data.data.physicalItemType "consumable")}} ({{data.data.quantity}}){{/if}}
             </div>
-            <div data-item-id="{{id}}" class="gear contextmenu nowrap" style="display: block;">
+            <div data-item-id="{{id}}" class="gear usagecontainer contextmenu nowrap">
               {{#each data.data.usage}}
                 {{#if skillId}}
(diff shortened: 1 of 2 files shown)


src/actors/sheet-parts/gear-tab.hbs
```diff
@@ -190,7 +190,7 @@
               {{/each}}
             </span>
             </div>
-            <div data-item-id="{{id}}" class="gear contextmenu">
+            <div data-item-id="{{id}}" class="gear contextmenu" style="display: block;">
```
wake42, line 193: see if you can modify the rqg.scss instead of hardcoding the styling

wake42 (changes requested):

Moonpile:
OH GOD this took me forever and gave me feelings of inadequacy!  I did however correct something beyond just the "usage" styles now being in css.  If the weapon only has one usage then the usage cells (Skill, Damage, STR, DEX) line up with the common non-usage cells (Name HP, Price, ENC, Location).  If the weapon has more than one usage cell then the non-usage cells "straddle" the usage cells to make it clear that they apply to the weapon overall and not just the usage, as shown below.

image

Moonpile:
OH GOD this took me forever and gave me feelings of inadequacy!  I did however correct something beyond just the "usage" styles now being in css.  If the weapon only has one usage then the usage cells (Skill, Damage, STR, DEX) line up with the common non-usage cells (Name HP, Price, ENC, Location).  If the weapon has more than one usage cell then the non-usage cells "straddle" the usage cells to make it clear that they apply to the weapon overall and not just the usage, as shown below.

image

wake42 (approved):

PR #110, opened 2021-12-22
feat Calc total worth in currency item

### LINKED ISSUE #96: Calc total worth in currency item
In the Gear tab - currency table the "Price" column should show quantity * item price.
There is a handlebar handler called multiply that could be used (and ENC is already using that).

image

Moonpile:
Fixes #96

- added column with extended value of coins
- created new "currency" grid class with 8 columns
- applied text-right to the quantity textbox

### DIFF
--- a/src/actors/sheet-parts/gear-tab.hbs
+++ b/src/actors/sheet-parts/gear-tab.hbs
@@ -95,13 +95,13 @@
                      size="10">
             </div>
             {{/each}}
-            <div class="foot1"></div>
-            <div class="foot2">Total</div>
-            <div class="foot3"><span class="text-right"></span></div>
-            <div class="foot4"><span class="text-right">{{numberFormat currencyTotals.price.estimated decimals=2}}&thinsp;L</span></div>
-            <div class="foot5"><span class="text-right">{{numberFormat currencyTotals.encumbrance decimals=2}}</span></div>
-            <div class="foot6"></div>
-            <div class="foot7" style="background-color: transparent;"></div>
+            <div class="no-background"></div>
+            <div class="no-background">Total</div>
+            <div class="no-background"><span class="text-right"></span></div>
+            <div class="no-background"><span class="text-right">{{numberFormat currencyTotals.price.estimated decimals=2}}&thinsp;L</span></div>
+            <div class="no-background"><span class="text-right">{{numberFormat currencyTotals.encumbrance decimals=2}}</span></div>
+            <div class="no-background"></div>
+            <div class="no-background" style="background-color: transparent;"></div>
             <div class="footer"></div>
           </div>
         </article>
--- a/src/actors/sheet-parts/gear-tab.hbs
+++ b/src/actors/sheet-parts/gear-tab.hbs
@@ -74,8 +74,7 @@
             <div class="head5"><span class="text-right">ENC</span></div>
             <div class="head6"></div>
             <div class="head7">Location</div>
-            {{#each ownedItems.gear}}
-            {{#if (eq data.data.physicalItemType "currency")}}
+            {{#each ownedItems.currency}}
             <div data-item-id="{{id}}" class="gear contextmenu"><img class="item" src="{{img}}"></div>
             <div data-item-id="{{id}}" class="gear contextmenu" title="{{data.data.price.conversion}}">{{name}}</div>
             <div data-item-id="{{id}}" class="gear contextmenu">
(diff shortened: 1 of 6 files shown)


Moonpile:
image

This was a good one for me to finally figure out how grids work!

Two possible enhancements from here:
1) make a Total row at the bottom with total wealth as currency
2) if it's too crowded in that grid, maybe make the "Price" and "Ext" (Extended) columns toggle back and forth into visibility when you click the header.

wake42 (changes requested):
> make a Total row at the bottom with total wealth as currency

That would be a really nice addition I agree

> if it's too crowded in that grid, maybe make the "Price" and "Ext" (Extended) columns toggle back and forth into visibility when you click the header.

Maybe we don't need to show the value (conversion factor) of a single coin. What do you think of only showing the multiplied value.

Moonpile:
> Maybe we don't need to show the value (conversion factor) of a single coin. What do you think of only showing the multiplied value.

I thought about just pulling out "Price" which is really conversion factor in this case but I really liked the idea of it being there for quick reference to the players.  I think either the click to swap "Ext" and "Price" idea or maybe just making the "Price" column a lighter color so the eye isn't drawn to it as much and it doesn't distract from the "Ext" which is what most people want to see most of the time.

wake42:
> I thought about just pulling out "Price" which is really conversion factor in this case but I really liked the idea of it being there for quick reference to the players. I think either the click to swap "Ext" and "Price" idea or maybe just making the "Price" column a lighter color so the eye isn't drawn to it as much and it doesn't distract from the "Ext" which is what most people want to see most of the time.

That might be it, remove the heading for the conversion factor and make the font smaller. Maybe add a × in front so you have the input field with quantity and then × 0.1 in a small font and the total after that in normal font size.

Moonpile:
Removed the Price column and made a tooltip over the name to tell the conversion factor in plain language.

Note that I cannot get the total footer's background div to go behind the rest of the footer divs the way the header works.
image

wake42:
Really nice! I have a half solution for getting the text into the footer. If you add `margin-bottom: -1.4rem` to the footer, the it will look like this:
image
The alternating background is still there though which isn't so nice... 

I also suggest that we remove the sum of Qty since that really doesn't add any interesting info.

I really like the tooltip with the currency conversion!

wake42:
One more thing to consider is how ENC rounding should work. I agree that having 2 decimals like that is nicer, but in the system it's simplified for human calculations so that if it's less than 1 then it's 0 (at least that's how I interpret it) It is a bit strange though that if you have 159 lunars then it's weightless - but add one more and it's 1 ENC, so we should figure out how to handle that (generally with ENC calculations). 

And a GUI issue - I think there should be a little more space between Ext and ENC columns to visually divide them a bit more.

Moonpile:
I fixed the footer, removed the total number of coins (which was fun but junking it up), added some space between Ext and ENC, and sorted the coins by decreasing price.  

This does not address your concern about the ENC which it think merits a different issue to handle it everywhere.

wake42:
I had a look at the alternating shading and played around with the looks a bit. What do you think about not making the footer look like header, but just making it yellow like this:

image

I also realised that the formatting could be simplified, my footer looks like this 
```css
    .footer {
      grid-column: 1/-1;
      margin-top: -1.3rem;
      mix-blend-mode: color-burn;
      text-shadow: 3px 4px 4px var(--rqg-color-grid-text-shadow);
      box-shadow: var(--rqg-box-shadow-grid-header);
    }
```

and the foot1, foot2 etc isn't really needed. One can make a class like this and add that to each footer cell instead.
```css
.no-background {
      background-color: transparent !important;
    }
```

I usually don't like !important, but to achieve greater specificity than the alternating background I think would make a pretty ugly selector. Also this could be used in other places possibly.

There was a leftover `style="background-color: transparent;"` in gear-tab.hbs on the foot7 line104 that can be removed as well.

Moonpile:
I removed the foot# classes and added no-background.  I kind of like it better with the footer otherwise looking exactly like the header, but if you like it otherwise just change it.  I was wondering what you were talking about with the alternating color until I gave my demo actor an odd number of currencies.   I thought I had it solved until then!

src/actors/sheet-parts/gear-tab.hbs
```diff
@@ -69,22 +69,22 @@
             <div class="head1"></div>
             <div class="head2">Name</div>
             <div class="head3"><span class="text-right">Qty</span></div>
-            <div class="head4"><span class="text-right">Price</span></div>
+            {{!-- <div class="head4"><span class="text-right">Price</span></div> --}}
```
wake42, line 72: Leftover comment

src/rqg.scss
```diff
@@ -272,6 +276,20 @@
...
 
+    .footer {
+      grid-column: 1/-1;
+      grid-row: auto;
+      margin-top: -1.3rem;
+      background: var(--rqg-color-main-bg);
+      mix-blend-mode: color-burn;
+      text-shadow: 3px 4px 4px var(--rqg-color-grid-text-shadow);
+      box-shadow: var(--rqg-box-shadow-grid-header);
+      border: 3px double grey;
```
wake42, line 287: Could we meet halfway :-) What I'm after is differentiating headers and footer, but this is a little less differentiating than my previous suggestion: 
```suggestion
      border: 1px solid grey;
```
 
image

wake42 (changes requested):

src/rqg.scss
Moonpile: > Could we meet halfway :-) What I'm after is differentiating headers and footer, but this is a little less differentiating than my previous suggestion:

I'm no visual artist but that's the best solution so far, I think.  Looks similar but is obviously different.

wake42 (approved):

PR #112, opened 2021-12-22
feat Investigate default item img per type

### LINKED ISSUE #97: Investigate default item img per type
See if it's possible to define a default img for newly created items of a specific type. For example give all runeMagic items a img different from the standard "bag" icon.

There are proposed img for hitlocation / runeMagic & spiritMagic in systems/rqg/assets/images/items

Moonpile:
Fixes #97

- Update image for Hit Location, Rune Magic Spell, and Spirit Magic Spell on getData.
- This means you can't set those objects back to the default Foundry bag image.


Moonpile:
Two ideas for further enhancement:

- If you want to be able to set the image back to the default foundry bag image, we could put an initialized flag on the item and set it to true in a write-only fashion
- We could get the image URL from localization so that it is effectively a setting instead of hard-coded.

src/items/hit-location-item/hitLocationSheet.ts
```diff
@@ -49,6 +49,10 @@ export class HitLocationSheet extends RqgItemSheet<
     const itemData = this.document.data.toObject(false);
     assertItemType(itemData.type, ItemTypeEnum.HitLocation);
 
+    if (this.item.img === "icons/svg/item-bag.svg") {
```
wake42, line 52: The foundry default icon can be found in this variable `foundry.data.ItemData.DEFAULT_ICON`

wake42 (commented):
> If you want to be able to set the image back to the default foundry bag image, we could put an initialized flag on the item and set it to true in a write-only fashion

I think it's fine not to be able to reset to bag icon. If we really wanted I guess we could change the system standard `foundry.data.ItemData.DEFAULT_ICON` to a copy of the bag icon and check against that. Then it would be possible to change to then "normal" system standard bag if you wanted to. 
Actually gearItem might need a bag icon, so maybe we need to make our own bag anyway. 

> We could get the image URL from localization so that it is effectively a setting instead of hard-coded.

Maybe we should make it a proper setting, and why not include all the itemTypes then. I think there are a filepicker for settings so it should be easy to change for the user.

Moonpile:
Reworked this entirely by overriding onCreate in rqgItem to set default icons and when items have more complicated names to make sure that the user inputted name from the Foundry create item dialog gets somewhere meaningful.  Also provided system settings for default icons and localizations for those settings.

**Some new icons pulled from game-icons-net module**.  I documented which ones they were in the commit.

src/items/rqgItem.ts
```diff
@@ -96,6 +100,101 @@ export class RqgItem extends Item {
     });
   }
 
+  protected _onCreate(
+    data: ItemDataSource,
+    options: DocumentModificationOptions,
+    userId: string
+  ): void;
```
wake42, line 107: This is just duplicated code I think?

src/items/rqgItem.ts
```diff
@@ -96,6 +100,101 @@ export class RqgItem extends Item {
...
+  ): void {
+
+    if (data.type === ItemTypeEnum.Armor) {
+      //@ts-ignore
+      const item = getGame().items?.get(data._id);
+      item?.update({
+        img: String(getGame().settings.get("rqg", "defaultIconArmor")),
+        "data.namePrefix": data.name,
+      });
+    }
```
wake42, line 121: If we stored _one_ defaultIcons setting with an object like
```ts
{
[ItemTypeEnum.Armor]: "systems/rqg/assets/images/armor/cuirass.svg",
[ItemTypeEnum.Cult]: "systems/rqg/assets/images/items/cult.svg",
...
}
```
Then this repetition could be simplified to something like

```ts
const defaultIcons = getGame().settings.get("rqg", "defaultIcons"));
//@ts-ignore
const item = getGame().items?.get(data._id);
item?.update({
  img: defaultIcons[data.type],
});
```
And the settings for icons would be grouped into a separate dialog. They are quite many and overpowers the system settings  a bit. This would require creating a default icons settings Dialog though. I haven't looked in to how complicated that is...

The (very good) fix for item names would still require an "if" I guess for the item types that need it.

src/items/rune-magic-item/runeMagicSheet.ts
```diff
@@ -47,6 +47,10 @@ export class RuneMagicSheet extends RqgItemSheet<
     const itemData = this.document.data.toObject(false);
     assertItemType(itemData.type, ItemTypeEnum.RuneMagic);
 
+    // if (this.item.img === "icons/svg/item-bag.svg") {
+    //   this.item.update({ img: "/systems/rqg/assets/images/items/rune-magic.svg" }, {});
+    // }
```
wake42, line 52: Leftover commented code.

src/items/spirit-magic-item/spiritMagicSheet.ts
```diff
@@ -39,6 +39,10 @@ export class SpiritMagicSheet extends RqgItemSheet<
     const itemData = this.document.data.toObject(false);
     assertItemType(itemData.type, ItemTypeEnum.SpiritMagic);
 
+    // if (this.item.img === "icons/svg/item-bag.svg") {
+    //   this.item.update({ img: "/systems/rqg/assets/images/items/spirit-magic.svg" }, {});
+    // }
```
wake42, line 44: Leftover commented code

wake42 (changes requested):

src/items/rqgItem.ts
Moonpile: VS code autogenerated it this way with both signatures.  Removing the one you pointed out here still works.  I guess I didn't know why it was doing that.

src/items/rune-magic-item/runeMagicSheet.ts
Moonpile: Removed

src/items/spirit-magic-item/spiritMagicSheet.ts
Moonpile: Removed

src/items/rqgItem.ts
Moonpile: I agree that it junks up the settings.  I couldn't find some way to launch a custom dialog from the settings.  Or would we have to have some other way to launch that dialog?  Or are you suggesting just one text setting with a JSON blob in it?

wake42: I haven't looked into it much myself, but this describes how to set it up. https://foundryvtt.wiki/en/development/api/settings#setting-menus
I think the actual setting can be an Object and the setting FormApplication would just show a list of filepickers - one for each itemtype.

Moonpile:
I've got the settings as one object and laid the groundwork for the dialog to change them but that's going to take some more puzzling over.

src/dialog/improveAbilityDialog.ts
```diff
@@ -8,7 +8,7 @@ import { assertItemType, getGame } from "../system/util";
 export async function showImproveAbilityDialog(
   actor: RqgActor,
   itemId: string,
-  item: Item,
+  item: RqgItem,
```
Moonpile, line 11: I merged master into issue97 earlier today and this wasn't building in issue97, though it built fine in master as "Item".  I believe the problem was at line 72 "await submitImproveAbilityDialog(...." where if item is Item it wasn't working but if item is RqgItem it works fine. I think it actually makes sense that it's an RqgItem and it builds and works that way.  This is not related to the default icons issue.  Not sure if this is pointing out some other issue or not.

src/dialog/DefautIconSettings.ts
```diff
@@ -0,0 +1,32 @@
...
+      id: "default-icons-settings-dialog",
+      title: "Default Item Icon Settings",
+      template: "./systems/rqg/dialog/defaultIconSettings.hbs",
+      classes: ["form", "default-icons"],
+      width: 500,
+      closeOnSubmit: false,
+      submitOnChange: true,
+    });
+  }
+
```
wake42, line 20: Adding this will update the img icon immediately when selecting an image file. And then the OK button isn't really neccesary (since it doesn't actually do anything).
```suggestion
  protected _onSelectFile(selection: string, filePicker: FilePicker): void {
    super._onSelectFile(selection, filePicker);
    this.submit();
  }
```
At some point one might want to have a save / cancel button there, but that sounds like unneccesary work for now.

wake42 (changes requested):

wake42 (approved):

PR #113, opened 2021-12-22
feat Implement improvement for Passions, Runes, and Characteristics

Moonpile:
- Moves the existing skill improvement dialog to a generic dialog for all improvement.
- Allows for experience and training for Skills
- Allows for experience for Passions
- Allows for experience and training for Runes
- Allows for experience and training for POW
- Allows for training and research for STR, CON, DEX, and CHA


src/dialog/improveAbilityDialog.ts
```diff
@@ -0,0 +1,594 @@
...
+      img: item.img,
+      chance: ability.chance || 0,
+      chanceToGain: 100 - Number(ability.chance) || 0,
+      experienceGainFixed: 3,
+      experienceGainRandom: "1d6",
+      trainingGainFixed: 2,
+      trainingGainRandom: "1d6-1",
+    };
+
+    if (item.type === "passion") {
```
wake42, line 32: The typescript typing of foundry likes it better if you check `item.data.type` rather than `item.type` even if they are equivivalent. Also you can check against the Enum instead of just a string. This is in a couple of places, but I only comment on this one.
```suggestion
    if (item.data.type === ItemTypeEnum.Passion) {
```

src/dialog/improveAbilityDialog.ts
```diff
@@ -0,0 +1,594 @@
...
+import { RqgActor } from "../actors/rqgActor";
+import { Characteristic } from "../data-model/actor-data/characteristics";
+import { ItemTypeEnum } from "../data-model/item-data/itemTypes";
+import { RuneDataProperties } from "../data-model/item-data/runeData";
+import { Ability, IAbility } from "../data-model/shared/ability";
+import { RqgItem } from "../items/rqgItem";
+import { assertItemType, getGame } from "../system/util";
+
+export class ImproveAbilityDialog {
+  //**Shows a dialog for improving a Passion, Rune, or Skill */
```
wake42, line 11: I think we can skip the class and export the functions directly. I'm not consistent in the system, and when I started I leaned against making classes for things (coming from Java / Scala) but I now think that it doesn't add anything - at least not if the class doesn't store something as well. It's definitely a matter of preference, but I would say that we should not use "function holding classes".

wake42 (commented):
I like this approach! I would however split the improveAbilityDialog.ts file into two separate, one for Improve Ability and one for Improve Characteristic. They can both use the same hbs template I think, but that file is too long for it's own good.

src/dialog/improveAbilityDialog.ts
Moonpile: done

src/dialog/improveAbilityDialog.ts
Moonpile: done

wake42 (approved):

Issue #138, opened 2021-12-29
Make chat cards have a unified look/feel


Moonpile:
This should take into account chat types like "whisper" vs "public", etc.  And it should try to resemble the newer books.

### Related
- [x] #526

wake42:
Part of a big bang PR

Issue #123, opened 2021-12-25
Layout of rune tab


wake42:
The layout of the rune tab is currently hard coded to the rune names. This doesn't work if the names are translated. It also works badly if not all runes are there since opposed runes look bad if the opposed is missing.

Loose ideas:
It would be nice if the runes were drawn from the opposed / minor data...
- html på canvas - draw lines between input fields   http://jsfiddle.net/stevendwood/5sSWj/ 
- Math for polygons like pentagram 🙂  https://codepen.io/thebabydino/pen/ybVEzP/?editors=1100

Moonpile:
If we can get it into something like this then being sorted by name after localization doesn't matter.  I'm not sure how to go about this in html/css though:

![image](https://user-images.githubusercontent.com/3959760/153211556-9c3ffe54-a4dc-4599-95e5-ee0aedd36675.png)

wake42:
I also had another idea to fix this. If we let the item name be translated to whatever, but require an english name as well, then the actual item name will be shown in the compendiums etc, but we could use the data.en-name to know which rune it is.
Haven't thought this through though...

wake42:
Problem of referencing runes would be solved by #199.
Leaving only to solve the layout if some runes are missing.

Moonpile:
Also keep in mind that Actors might get "extra runes that don't fit in the layout shown above.  Like if you gain the Infinity rune or Dragon rune, or if you're an elf and have Plant <> Man in opposition.  I think a list of any extra runes below this layout would be fine.

The Man <> Beast Form Rune opposition going through the Power Rune Affinities is cool, but it would be neat if we could make it so that you could change the runes in the layout, so an elf could have Plant <> Man there instead, for example.

Also Illumination allows you to unlink opposed Runes.  It think it would be cool if those runes stayed where they were and the line between them simply went away.

And NPCs might not have the full compliment of runes.  It might be nice to have just a simple "List View" as an option that you can switch to instead of the full layout.

wake42:
Don't look too hard at the screenshot above, the starter set runes have a different look. And I think we should implement something that works well on screen – not blindly copying the paper character sheets. And yes, the layout should be able to dynamically handle "missing" runes as well as different combinations like man<->beast &  man<->plant etc.
![image](https://user-images.githubusercontent.com/4853677/216074770-b686a98b-625a-4ed0-8bf3-913291ebf290.png)

PR #143, opened 2021-12-30
feat Create a RuneMagic Chat Card

Moonpile:
Created runeMagicCard that displays in chat when player clicks a Rune Magic Spell on the character sheet.  It has a link to the journal description and allows the player to select the rune to cast with, rune points to spend, and magic points to boost.  It also allows player to augment with ritual or meditation, skill augmentation, and other modifiers.  Clicking roll resolves the action and removes Rune Points and Magic Points if needed.

In the runeMagicCard.ts there are a number of //@ts-ingores that we might be able to clean up.  Also I'm concerned about "_id" usage. 

The look and feel of the card is not meant to be final, but just to indicate the two "groups" and to suggest a direction for future work no card look/feel in issue #138


src/_variables.scss
```diff
@@ -25,6 +25,8 @@
   --rqg-color-enc-legend: #e4cc9d;
 
 
+  --rqg-color-invalid-joural: #e4b5ad; //Regular journal in chat card is: #7a7971
```
wake42, line 28: Misspelled ...-journal

src/chat/runeMagicCard.ts
```diff
@@ -0,0 +1,360 @@
...
+    token: TokenDocument | null
+  ): Promise<void> {
+    const runeMagicItem = actor.items.get(runeMagicItemId);
+    assertItemType(runeMagicItem?.data.type, ItemTypeEnum.RuneMagic);
+    const cult = actor.items.get(runeMagicItem.data.data.cultId);
+    assertItemType(cult?.data.type, ItemTypeEnum.Cult);
+
+    let usableRuneNames: string[] = [];
+    let runesForCasting: RuneDataSource[] = [];
+    if (runeMagicItem.data.data.runes.includes("Magic (condition)")) {
```
wake42, line 59: I don't have any solution and I do the same thing somewhere else, but hardcoding the rune name makes it hard to translate. I have special skill names as [config](https://github.com/sun-dragon-cult/fvtt-system-rqg/blob/160aae27f89fd51e1a56a572f72dc206cda8a03b/src/system/config.ts#L22) - could this special rune possibly be handled the same way?

src/chat/runeMagicCard.ts
```diff
@@ -0,0 +1,360 @@
...
+    }
+  }
+
+  private static async GiveExperienceToRune(
+    actor: RqgActor,
+    runeId: string
+  ) {
+    await actor.updateEmbeddedDocuments("Item", [
+      { _id: runeId, data: { hasExperience: true } },
+    ]);
```
wake42, line 298: Getting exp. is nice - we should probably do a ui.notification as well to make the player realise what happened.

src/actors/rqgActorSheet.ts
```diff
@@ -738,6 +739,54 @@ export class RqgActorSheet extends ActorSheet<
       });
     });
 
+    // Roll Rune Magic
+    this.form?.querySelectorAll("[data-rune-magic-roll]").forEach((el) => {
```
wake42, line 743: add `[data-rune-magic-roll]`to rqg.scss around line 562 to get a hover effect

wake42 (changes requested):
Also the right click context menu for runemagic needs to be updated - it still uses skillRolls.

src/_variables.scss
Moonpile: Yeah, I noticed that too.  Done.

src/chat/runeMagicCard.ts
Moonpile: Made the "magicRuneName" setting and used it.

src/actors/rqgActorSheet.ts
Moonpile: Done

src/chat/runeMagicCard.ts
Moonpile: Created Actor.AwardExperience.  Eventually all experience gain should funnel through this method for consistency.

src/chat/runeMagicCard.ts
```diff
@@ -0,0 +1,411 @@
...
+    // Get the actor's versions of the runes, which will have their "chance"
+    usableRuneNames.forEach((rune) => {
+      //@ts-ignore name
+      const actorRune = actor.items.getName(rune);
+      assertItemType(actorRune?.data.type, ItemTypeEnum.Rune);
+      runesForCasting.push(actorRune.data);
+    });
+
+    const strongestRune = runesForCasting.reduce(function (prev, current) {
+      //@ts-ignore data WHY?!
```
wake42, line 78: Don't think this ts-ignore is needed?

src/chat/runeMagicCard.ts
```diff
@@ -0,0 +1,411 @@
...
+      // And some cults have the same rune more than once, so de-dupe them
+      usableRuneNames = [...new Set(cult.data.data.runes)];
+    } else {
+      // Actor can use any of the Rune Magic Spell's runes to cast
+      usableRuneNames = [...new Set(runeMagicItem.data.data.runes)];
+    }
+
+    // Get the actor's versions of the runes, which will have their "chance"
+    usableRuneNames.forEach((rune) => {
+      //@ts-ignore name
```
wake42, line 71: Don't think this ts-ignore is needed?

src/chat/runeMagicCard.ts
```diff
@@ -0,0 +1,411 @@
...
+    actor: RqgActor,
+    speakerName: string
+  ): Promise<void> {
+    assertItemType(runeMagicItem?.data.type, ItemTypeEnum.RuneMagic);
+    const cult = actor.items.get(runeMagicItem.data.data.cultId);
+    assertItemType(cult?.data.type, ItemTypeEnum.Cult);
+    
+    let usableRuneNames: string[] = [];
+    let runesForCasting: RuneDataSource[] = [];
+    if (runeMagicItem.data.data.runes.includes(getGame().settings.get("rqg", "magicRuneName") as string)) {
```
wake42, line 198: The settings can be typed [here](https://github.com/sun-dragon-cult/fvtt-system-rqg/blob/1400dec6fdde2d9b39b2b1eb86cfd2ac59dc9155/src/global.d.ts#L30) so that no `as string` is needed.

src/chat/runeMagicCard.ts
```diff
@@ -0,0 +1,411 @@
...
+      // And some cults have the same rune more than once, so de-dupe them
+      usableRuneNames = [...new Set(cult.data.data.runes)];
+    } else {
+      // Actor can use any of the Rune Magic Spell's runes to cast
+      usableRuneNames = [...new Set(runeMagicItem.data.data.runes)];
+    }
+
+    // Get the actor's versions of the runes, which will have their "chance"
+    usableRuneNames.forEach((rune) => {
+      //@ts-ignore name
```
wake42, line 209: the ts-ignore is not needed I think

src/chat/runeMagicCard.ts
```diff
@@ -0,0 +1,411 @@
...
+      const actorRune = actor.items.getName(rune);
+      assertItemType(actorRune?.data.type, ItemTypeEnum.Rune);
+      runesForCasting.push(actorRune.data);
+    });
+
+    const strongestRune = runesForCasting.reduce(function (prev, current) {
+      //@ts-ignore data WHY?!
```
wake42, line 216: the ts-ignore is not needed I think

src/i18n/en.json
```diff
@@ -238,7 +242,61 @@
...
+      "SkillAugmentationValue30": "Special Success, +30%",
+      "SkillAugmentationValue20": "Success, +20%",
+      "SkillAugmentationValue-20": "Failure, -20%",
+      "SkillAugmentationValue-50": "Fumble, -50%",
+      "OtherModifiers": "Other Modifiers:",
+      "getActorFromIdsWarning": "Couldn't get actor from actor id or token id.",
+      "noSelectedRuneWarning": "Attempt to roll a Rune Magic Spell with no Rune selected on the card.",
+      "resultMessageCritical": "The spell takes effect and costs no Rune Points and {magicPointBoost} Magic Points from boosting.",
+      "resultMessageSpecial": "The spell takes effect and costs {runePointCost} and {magicPointBoost} Magic Points from boosting.",
+      "resultMessageSuccess": "The spell takes effect and costs {runePointCost} and {magicPointBoost} Magic Points from boosting.",
```
wake42, line 286: Maybe add "rune points" to make it more clear what the cost is
```suggestion
      "resultMessageSpecial": "The spell takes effect and costs {runePointCost} Rune Points and {magicPointBoost} Magic Points from boosting.",
      "resultMessageSuccess": "The spell takes effect and costs {runePointCost} Rune Points and {magicPointBoost} Magic Points from boosting.",
```

wake42 (changes requested):

src/chat/runeMagicCard.ts
wake42: Nice! The only exception would be the POW exp check I think since everything else is items. Maybe add a case for refactoring the other experience checks.

src/chat/runeMagicCard.ts
wake42: That's much better than the config I have for the skills. I'll add a case for refactoring the special skills into a setting as well. I guess all of those could become a FormApplication for setting "special" items names. It possibly ties in a little with #123 as well (even though I don't want all rune names to be settings)

src/chat/runeMagicCard.ts
Moonpile: Weird.  Because I definitely double-checked all the ts-ignores and couldn't get rid of them before, but now I can.

Moonpile: same

src/chat/runeMagicCard.ts
Moonpile: I figured there must be a way to type that!

wake42 (commented):
Looks good, the `--rqg-color-invalid-joural` is still misspelled.
And I think we can make another case for supporting one-use spell cost, or should it go into this one as well?

src/_variables.scss
Moonpile: I might have put this reply on the wrong comment earlier!

Moonpile:
fixed the `--rqg-color-invalid-joural`

I suppose we should do the one-use at this point too, since it's deeply bound into all of this other work here.

wake42 (approved):

PR #145, opened 2021-12-31
Remove linked RuneMagic when a Cult is deleted

### LINKED ISSUE #128: Remove linked RuneMagic when a Cult is deleted
When a character leaves a cult it should retain the cult and all it's linked RuneMagic. Leaving a cult just means you can't replenish your Rune Points.

Deleting a Cult from an Actor on the other hand is more configuration and the RuneMagic that is linked to the Cult should also be removed then.

Moonpile:
Fixes #128

- rqgActorSheet.confirmItemDelete now checks if the item is a Cult and if so displays a different message and if accepted deletes associated Rune Magic Spells.
- localization
- !!! rqgItem.updateDocuments required changes to ignore empty updates.  Is this going to affect other things?
- rqgItem.localizeItemTypeName now handles localizing values from ItemTypeEnum
- the "localizeitemtype" handlebars helper now uses rqgItem.localizeItemTypeName

### DIFF
--- a/src/i18n/en.json
+++ b/src/i18n/en.json
@@ -250,9 +250,18 @@
       "CurrencyConversionTipOver1": "One {name} is worth {value} Lunars.",
       "CurrencyConversionTipLunar": "Lunars are the standard basis of currency.",
       "CurrencyConversionTipUnder1": "One Lunar is worth {value} {name}s."
+    },
+    "confirmItemDeleteDialog": {
+      "title": "Delete {itemType}: {itemName}",
+      "content": "Are you sure you want to delete the {itemType} named {itemName}?",
+      "contentCult": "Are you sure you want to delete the {itemType} named {itemName}? This will also delete all of the {runeMagicSpell} items associated with this {itemType}."
     }
   },
   "DIALOG": {
+    "Common": {
+      "btnCancel": "Cancel",
+      "btnConfirm": "Confirm"
+    },
     "improveAbilityDialog": {
       "title": "Improvement for {typeLocName} \"{name}\"",
       "titleChar": "Improvement for Characteristic \"{name}\"",
--- a/src/items/rqgItem.ts
+++ b/src/items/rqgItem.ts
@@ -121,22 +121,26 @@ export class RqgItem extends Item {
     const { parent, pack, ...options } = context;
     if (parent?.documentName === "Actor") {
       updates.forEach((u) => {
-        const document = parent.items.get(u._id);
-        if (!document || document.documentName !== "Item") {
-          const msg = "couldn't find item document from result";
-          ui.notifications?.error(msg);
-          throw new RqgError(msg, updates);
+        if (u && u.length > 0) {
+          const document = parent.items.get(u._id);
+          if (!document || document.documentName !== "Item") {
+            const msg = "couldn't find item document from result";
+            ui.notifications?.error(msg);
+            throw new RqgError(msg, updates);
+          }
+          // Will update "updates" as a side effect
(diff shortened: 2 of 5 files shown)


wake42:
Just a quick look, but adding runemagic to a cult is broken. 
I dragged a cult to new actor and then dragged a runemagic to the same actor and the runemagic is no longer linked to the cult.

src/items/rqgItem.ts
```diff
@@ -121,22 +121,26 @@ export class RqgItem extends Item {
...
-        // Will update "updates" as a side effect
-        ResponsibleItemClass.get(document.data.type)?.preUpdateItem(
-          parent,
-          document,
-          updates,
-          options
-        );
       });
     }
-    return super.updateDocuments(updates, context);
```
Moonpile, line 139: The problem is here, but it's going to throw an exception when you delete a cult unless you do the if (updates[0].length > 0) check, BUT, for some reason, if you do that check then it doesn't add new spells to the cult.

src/i18n/en.json
```diff
@@ -250,9 +250,18 @@
       "CurrencyConversionTipOver1": "One {name} is worth {value} Lunars.",
       "CurrencyConversionTipLunar": "Lunars are the standard basis of currency.",
       "CurrencyConversionTipUnder1": "One Lunar is worth {value} {name}s."
+    },
+    "confirmItemDeleteDialog": {
```
wake42, line 254: Maybe this should be called `confirmCultDeleteDialog` since `confirmItemDeleteDialog` could be used for a more generic delete dialog.

src/system/registerHandlebarsHelpers.ts
```diff
@@ -25,7 +26,8 @@ export const registerHandlebarsHelpers = function () {
   });
 
   Handlebars.registerHelper("localizeitemtype", (typeName) => {
-    return getGame().i18n.localize("ITEM.Type" + typeName.titleCase());
+    const itemType: ItemTypeEnum = <ItemTypeEnum>typeName;
```
wake42, line 29: `<ItemTypeEnum>typeName` is the older version on typecasting in ts. Now they recommend `typeName as ItemTypeEnum` but is the typecasting really necessary here?

src/items/rqgItem.ts
```diff
@@ -160,4 +167,10 @@ export class RqgItem extends Item {
     const okToAdd = !isRuneMagic || !(isRuneMagic && !actorHasCult);
     return !okToAdd;
   }
+
+  //** Localizes Item Type Name using ITEM localization used by Foundry.
+  //** See also  */
```
wake42, line 172: see also what?

wake42 (changes requested):

src/i18n/en.json
Moonpile: If you look in rqgActorSheet.confirmItemDelete you'll see that it actually IS the generic item delete, but it uses the "contentCult" for the content if it's a cult and otherwise just uses "content".   So I think "confirmItemDeleteDialog" makes sense.

src/system/registerHandlebarsHelpers.ts
Moonpile: Does not appear to be needed.

src/items/rqgItem.ts
Moonpile: Removed.  I was trying to cross-reference to the RqgItem to indicate that they both did a similar localization but then I had the better idea of just using the RqgItem localization in the handlebar helper.

wake42 (approved):

PR #147, opened 2022-01-03
Importing items from compendium clears image

wake42:
I now check if the item image is foundry default before setting RQG default image


Moonpile:
After our discussion I was able to reproduce the original problem in Foundry 9, and I'm satisfied with this fix.

Issue #105, opened 2021-12-21
Restructure translation keys


wake42:
All translation keys should be namespaced with RQG (with the exception of ITEM that is a foundry namespace).

Below RQG namespace the structure should be `<sheet/page/dialog>.<group-specificdialog>.key`

For example `RQG.ActorSheet.CombatTab.someKey` if the sheet/dialog is simple it's fine to skip the second grouping for example
`RQG.PassionSheet.someKey`

There are also exception for groupings that can be deeper, like `RQQ.ActorSheet.TopArea.Characteristic.strength` and for translations that appear in multiple places we can have `RQG.Common.CultRank.initiate`

Moonpile:
- Item (Foundry's default place to look for the localizable names of items)
- RQG
-- Settings
-- characteristic (merge with characteristicfull) ie:
---"str": "STR"
---"strength": "Strength"
-- ResultEnum

wake42:
A few thoughts... Right now we only have one actor "character" but we have talked about adding some sort of look actor etc. It will make the content keys longer, but should we have RQG.Actor.Character.xxx to future proof it?

Homeland and Occupation should probably be under background.

RQG.Game.CombatManeuver could be a problem since you can add whatever in that field (even if you get a bunch of suggested values). It might need some thinking of how to solve that.

Other than that it look good I think.

Moonpile:
I think if we have a loot actor, its localizations might go under "RQG.Actor.Loot.foo", or even just "RQG.LootActor.foo".

Agreed about Homeland and Occupation.

wake42:
I agree on the RQG.LootActor idea - no need to make the structure deeper than necessary.
And for the combat manoeuvres - possibly they too should become a settings panel where you could localise the preconfigured names. I think you could switch out the values of the settings if you switch language then (hopefully).
defaultDamageType (slash, impale, bash, ...) is localisable though since it's a defined set of values.

Moonpile:
I think ideas like the damage type will occur to us while we're doing this.  Do you think we're ready to start on this one?

wake42:
Yes, this is a case that it good to do sooner rather than later to minimise rewrites of coming translation keys.

PR #149, opened 2022-01-03
feat Click to roll against Reputation

Moonpile:
- Reputation value is now "text-right"
- "reputation.svg" came from "beams-aura.svg"

Still needs:
- localization
- Add Reputation to icons dialog?


Moonpile:
How do we want to make it obvious that Reputation is rollable?  Right now I just have the word "ROLL" there, but I figure it ought to look more like skills, but I couldn't figure out how to get the hover over shadow effect.

image

Also I've got the "other modifiers" localization from the RuneMagicCard (which I currently don't have merged).  I suppose this could go to some "common" localization group.

image

And I made that Reputation icon a setting like the item icons even though Reputation isn't an item.  I see no reason not to add it to that dialog, do you?  And maybe putting it on the background tab might help make it clearer that Reputation is rollable?

wake42:
I agree reputation can be handled as a "fake" item. It will be a bit strange to give it an `ITEM.TypeReputation` translation key
image
But maybe we can live with that.

To get the hover effect you just have to add the roll here https://github.com/sun-dragon-cult/fvtt-system-rqg/blob/160aae27f89fd51e1a56a572f72dc206cda8a03b/src/rqg.scss#L569 
I see that the runeMagic roll should be added as well. And I also see that I've forgotten to remove `[data-melee-roll]`, maybe you could sneak that change in while you edit it. It was for the old meleeWeapon items.
I'm uncertain if it would be enough affordance to be able to understand that it is clickable, but maybe together with that icon it could be enough, even without "ROLL". On the player side the whole text `[*] Reputation 20%` including the icon should be clickable I think 
image
On the GM side only `[*] Reputation` should be clickable (to make it possible to edit the value without automatically rolling :-)

I have been toying with the idea to put reputation on the top part of the character sheet. It is a defining number of the actor, but it's probably too crowded there already...

Moonpile:
Fixed the hover over covering the Reputation input and added Reputation to ITEM localization.

I think we agreed that the localization can await the en.json re-org.  It'll probably become clear to me where it makes sense for that to go.

image

src/actors/rqgActorSheet.ts
```diff
@@ -694,6 +695,33 @@ export class RqgActorSheet extends ActorSheet<
...
+            this.actor,
+            speakerName
+          );
+          clickCount = 0;
+        } else if (clickCount === 1) {
+          await ReputationCard.show(
+            this.actor,
+            // @ts-ignore wait for foundry-vtt-types issue #1165 #1166
+            this.token
+          );
```
wake42, line 720: There needs to be a setTimeout around this, otherwise it will always show a chatcard as well. Have a look at the other listeners.

src/actors/rqgActorSheet.ts
```diff
@@ -694,6 +695,33 @@ export class RqgActorSheet extends ActorSheet<
...
+      let clickCount = 0;
+      el.addEventListener("click", async (ev: Event) => {
+        clickCount = Math.max(clickCount, (ev as MouseEvent).detail);
+
+        if (clickCount >= 2) {
+          // @ts-ignore wait for foundry-vtt-types issue #1165 #1166
+          const speakerName = this.token?.name || this.actor.data.token.name;
+          await CharacteristicCard.roll(
+            "Reputation",
+            this.actor.data.data.background.reputation || 0,
```
wake42, line 708: For some reason I always get 0% chance when doubleclicking to do a direct roll.

wake42 (changes requested):

src/actors/rqgActorSheet.ts
Moonpile: Done

src/actors/rqgActorSheet.ts
Moonpile: I wasn't getting quite that behavior, but I gave ReputationCard a "directroll" method which sets up flags as if you just clicked Roll on the card without making any changes.  This works the same way as the RuneMagicCard.

wake42 (approved):
Approving, but with a comment (that can be addressed in #105).
The translation key for the reputation card modifier is `RQG.RuneMagicCard.OtherModifiers`. Is this a case where the key should be `RQG.Game.OtherModifiers` instead since it's used in more places than one?

PR #157, opened 2022-01-05
Rename Rune experience to hasExperience

wake42:
I've manually gone through all starter-set runes and changed it there. This migration should go through all world compendiums, world items, items embedded in actors both in world and in scenes.

Normally this will be run when there is a new release, to force run the migration you can run `game.system.api.migrate()` in the console (alternatively lower the current version setting to force a migration to start)

### DIFF
--- a/src/actors/rqgActorSheet.ts
+++ b/src/actors/rqgActorSheet.ts
@@ -1002,11 +1002,13 @@ export class RqgActorSheet extends ActorSheet<
                 console.log("CULT", item);
                 const cultId = item.id;
                 //@ts-ignore cultId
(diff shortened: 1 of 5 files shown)


wake42:
I just found a bug - context menu "toggle experience" doesn't seem to work. Will have a look at that...


image

wake42:
I've updated the check RqgItem row 146 to not check specifically for if updates[0].data exists. The update for toggle experience looked like `{data.hasExperience: true, _id: 'fOTX6poLtMzMnJQA'}` so the key `data` didn't exist. I now count the number of object keys in the update instead. If the update is an empty array as when deleting a cult [].length will be 0 so the update will not happen.

Moonpile (approved):

PR #158, opened 2022-01-06
Restructure translation keys

Moonpile:


Moonpile:
This is ready for review.  It's one of those ones where we made sweeping changes so I wouldn't be surprised if we missed a localization somewhere.  Also we have more work to do on localizing everything that is not already localized, so unless we catch something right now we have a lot of chances to find more.

I also think this is one of those things where you can come up with an organizational plan and then either regret it later or at least think of many alternatives later, but I hope it's an improvement that lets us move forward with the rest of the localizations.

src/actors/sheet-parts/background-tab.hbs
```diff
@@ -1,107 +1,109 @@
...
+  </fieldset>
+
+  <fieldset>
+    <legend>{{localize 'RQG.Actor.Background.Circumstance'}}</legend>
 
     <div class="nowrap">
       {{#if isGM}}
       <span data-reputation-roll>
-        <img src="{{defaultItemIconSrc "Reputation"}}" class="item">
+        <img src="{{defaultItemIconSrc " Reputation"}}" class="item">
```
wake42, line 104: Remove extra space
```suggestion
        <img src="{{defaultItemIconSrc "Reputation"}}" class="item">
```

src/actors/sheet-parts/background-tab.hbs
```diff
@@ -110,51 +112,69 @@
       </span>%
       {{else if data.data.background.reputation}}
       <span data-reputation-roll>
-        <img src="{{defaultItemIconSrc "Reputation"}}" class="item">
-        <b>{{localize 'RQG.Reputation'}}</b> {{data.data.background.reputation}}%
+        <img src="{{defaultItemIconSrc " Reputation"}}" class="item">
```
wake42, line 115: Remove extra space
```suggestion
        <img src="{{defaultItemIconSrc "Reputation"}}" class="item">
```

wake42 (changes requested):
Just remove those 2 spaces that made the reputation icon not show up and you can merge after that. It's looking good!

src/actors/sheet-parts/background-tab.hbs
Moonpile: AAAARGH!!!!!!  The "Format Document" does that.  I have to remember to either not format the document or remember to fix it.  Pain in the butt.

Moonpile: Done

src/actors/sheet-parts/background-tab.hbs
Moonpile: Done

wake42 (approved):

PR #162, opened 2022-01-07
feat Go through all texts and include in translation system

Moonpile:


src/i18n/en.json
```diff
@@ -115,14 +117,81 @@
...
+        "TotalHitPoints": "Total Hit Points",
+        "HealingRatePerWeek": "Healing rate {healingRate}/week",
+        "HitLocation": "Hit Location",
+        "HitPointsAbbr": "HP",
+        "ArmorPointsAbbr": "AP",
+        "Wounds": "Wounds",
+        "HealWound": "Heal Wound",
+        "AddWound": "Add Wound"
+      },
+      "Occupation": {},
```
wake42, line 167: Duplicate of Actor.Background.Occupation ?

src/i18n/en.json
```diff
@@ -344,9 +663,33 @@
...
+      },
+      "ContentLink": {
+        "CompendiumNameTitle": "Compendium [{compendiumName}]",
+        "WorldJournalEntry": "World Journal Entry",
+        "DropToConnectJournal": "Drop a journal entry on this item to connect a description.",
+        "DropToReplaceJournal": "Drop a journal entry on this item to replace the description."        
+      }
+    },
     "Game": {
       "ComabatManeuver": {},
```
wake42, line 690: Misspelled and unused?

src/i18n/en.json
```diff
@@ -344,9 +663,33 @@
...
       "DamageRollType": {},
```
wake42, line 691: Unused?

wake42 (changes requested):
Fantastic job! I thought the easiest way of finding missed translations was to make a "swedish" translation where the only change was to add an apple as the first character. Here are some screenshots of where the apple is missing. There are very few places not translated now!

image

image

image

image


image

I was thinking - for things like "cancel" maybe one could use foundry's translation key?
image

image


image

image

image

image

image

image

image

wake42:
I had a look at the extra white svgs for context menus and found a way to be able to use the normal black ones. One can put the css style `filter: invert(1);` on them and they will become black.

src/chat/reputationCard.ts
```diff
@@ -64,7 +64,7 @@ export class ReputationCard {
       const speakerName = getSpeakerName(flags.actorId, flags.tokenId);
       await ReputationCard.roll(flags, actor, speakerName);
     } else {
-      ui.notifications?.warn("Couldn't find world actor to do reputation roll");
+      ui.notifications?.warn(localize("RQG.Item.Notifications.CouldNotFindActorByIdsWarn", {actorId: flags.actorId, tokenId: flags.tokenId}));
```
wake42, line 67:
```suggestion
      ui.notifications?.warn(localize("RQG.Item.Notification.CouldNotFindActorByIdsWarn", {actorId: flags.actorId, tokenId: flags.tokenId}));
```

src/chat/itemCard.ts
```diff
@@ -120,7 +122,7 @@ export class ItemCard {
       const speakerName = getSpeakerName(flags.actorId, flags.tokenId);
       await ItemCard.roll(flags.itemData, modifier, actor, speakerName);
     } else {
-      ui.notifications?.warn("Could not find world actor to do ability roll");
+      ui.notifications?.warn(localize("RQG.Item.Notifications.CouldNotFindActorByIdsWarn", {actorId: flags.actorId, tokenId: flags.tokenId}));
```
wake42, line 125:
```suggestion
      ui.notifications?.warn(localize("RQG.Item.Notification.CouldNotFindActorByIdsWarn", {actorId: flags.actorId, tokenId: flags.tokenId}));
```

wake42 (changes requested):

src/chat/reputationCard.ts
Moonpile: This looks like it's done.

src/chat/itemCard.ts
Moonpile: Again, looks like I did this already..  I think I had the same thought before I even saw this.

wake42 (changes requested):
I feel bad for pointing to stuff here, this is such a monster task and I've complicate things by adding things before this PR is merged - sorry about that! It's really not much left now though:

Combat Tab & Header
image

Gear Tab - view names
image

Spirit Magic Tab - merge error extra columns
image
should look like image

Spirit Magic Sheet - new flags & merge error
image

Weapon Sheet - damage types
image

Apart from that I think it should be possible to use the "normal" svg runes by adding a css style `filter: invert(1);` That should make a black icon white, and by doing that we don't need duplicated white runes or the rune font I think.

Great work, and again sorry for the merge problems. I'll try to keep out now :-)

Moonpile:
I think that addresses the latest screenshots.  One thing I think we could make clearer: if isMatrix means that the spell is held in a spell matrix maybe we should change the indicator and MatrixQ localizations to "held in matrix" and "Held in Matrix?", respectively.

src/i18n/en.json
```diff
@@ -277,11 +471,39 @@
...
         "hitLocationRoll": "Roll Hit location",
-        "Modifier": "Modifier"
+        "Modifier": "Modifier",
+        "CantFindSkillError": "Can't find skill with itemId [{skillId}] on actor {actorName} to show a weapon chat card.",
+        "CantFindActorWarn": "Can't find world actor to do action",
+        "UsedLastOfAmmoWarn": "You just used the last of your {projectileName}!",
+        "OutOfAmmoWarn": "You don't have any {projectileName} to {combatManeuverName}!",
+        "CantFindChatMessageError": "Can't find chat message",
+        "HitLocationRollFlavor": "Hit Location Roll",
+        "UnkownButtonInCardError": "Unknown button \"{actionButton}\" in weapon chat card",
```
wake42, line 481:
```suggestion
        "UnknownButtonInCardError": "Unknown button \"{actionButton}\" in weapon chat card",
```

src/i18n/en.json
```diff
@@ -316,22 +538,256 @@
...
+        }
+      },
+      "RuneMagic": {
+        "Range": "Range",
+        "Duration": "Duration",
+        "DurationEnum": {
+          "": "None",
+          "instant": "Instant",
+          "temporal": "Temporal",
+          "duration": "Duration",
```
wake42, line 688: "duration" doesn't exist anymore (is replaced by "special")

src/i18n/en.json
```diff
@@ -316,22 +538,256 @@
...
+        "MasteredSorceryQ": "Mastered (Sorcery)?",
+        "RuneType": {
+          "element": "Element",
+          "power": "Power",
+          "form": "Form",
+          "condition": "Condition",
+          "technique": "Technique"
+        }
+      },
+      "RuneMagic": {
```
wake42, line 681: Break out most of this to "Spell" since there is now an [interface](https://github.com/sun-dragon-cult/fvtt-system-rqg/blob/master/src/data-model/item-data/spell.ts) that is common to both Spirit, Rune (and eventually Sorcery) Magic.

src/i18n/en.json
```diff
@@ -316,22 +538,256 @@
...
+          "focused": "Focused",
+          "permanent": "Permanent",
+          "special": "Special"
+        },
+        "Concentration": "Concentration",
+        "ConcentrationEnum": {
+          "": "",
+          "active": "Active",
+          "passive": "Passive"
+        },
```
wake42, line 756: Remove and use "Spell" (see above)

src/i18n/en.json
```diff
@@ -316,22 +538,256 @@
...
+        "SpellFocus": "Spell Focus:",
+        "RitualQ": "Ritual?",
+        "EnchantmentQ": "Enchantment?",
```
wake42, line 759: Remove and use "Spell" (see above)

src/i18n/en.json
```diff
@@ -113,16 +139,97 @@
...
+        "Armor": "Armor",
+        "Consumables": "Consumables",
+        "Currency": "Currency",
+        "WeaponsAndShields": "Weapons & Shields",
         "CurrencyConversionTipOver1": "One {name} is worth {value} Lunars.",
         "CurrencyConversionTipLunar": "Lunars are the standard basis of currency.",
-        "CurrencyConversionTipUnder1": "One Lunar is worth {value} {name}s."
+        "CurrencyConversionTipUnder1": "One Lunar is worth {value} {name}s.",
+        "Name": "Name",
+        "EncumbranceAbbr": "ENC",
```
wake42, line 167: Should this be moved to Game? there is both `Actor.Gear.EncumbranceAbbr`, `Actor.Attributes.EncumbranceAbbr` and `Item.Gear.EncumbranceAbbr`.  Not sure though, and that would maybe imply a lot of other stuff should be moved "out". We could also live with multiple translations of the same thing...

wake42 (changes requested):
There is a `RQG.Item.Gear.LunarsAbbr` that is not used on background tab (base income & ransom).

`Dialog.improveAbilityDialog.gainchar` should be called `gainChar`

Spelling errors in keys under `Dialog.improveAbilityDialog`:
`experiencGainFixed`
`experiencGainRandom`

src/i18n/en.json
Moonpile: I'm going to fix the spelling in the key, but I'm not actually finding the key anywhere.

Moonpile: Nevermind, found it!

src/i18n/en.json
Moonpile: fixed

src/i18n/en.json
Moonpile: In general I was trying NOT to de-dupe translations for two reasons. First, it was confusing to remember where I put the "common" translations.  Second, I figured there might be some weird circumstance where the duplication actually needed different translations in different places in some other language because of the context.  If we don't think the second reason is a problem, then I think it was still easier to put the translations "in place" and then, seeing the structure we created, decide on how to create "common" keys.

Moonpile: Of course there were definitely exceptions where I felt more confident that one translation would always work and where a clear way of putting them in a common place suggested itself.  Again, I thought it better to just get it done and consider re-organization when we know what we have to re-organize.

src/i18n/en.json
Moonpile: Fixed

src/i18n/en.json
Moonpile: fixed

src/i18n/en.json
Moonpile: fixed,

src/i18n/en.json
Moonpile: Which is to say I didn't change this.

wake42 (approved):

PR #180, opened 2022-02-10
bug Spirit Magic Sheet not persisting journal links

### LINKED ISSUE #166: bug Spirit Magic Sheet not persisting journal links
If you drag a local or compendium journal entry onto the Spirit Magic Item Sheet, it shows the link, but if you refresh the game (F5), the link is not persisted.

image

After refresh:
image

Moonpile:
Fixes #166

- added _onDrop to RqgItemSheet.ts
- add localizeItemType to util.ts
- used localizeItemType in title of RqgItemSheet.ts
- used localizeItemType in improveAbilityDialog.ts
- sorted ITEM types in en.json
- added Shamanicability to ITEM in en.json
- added Sorcerymagic to ITEM in en.json
- removed _onDrop from cultSheet.ts, runeSheet.ts, runeMagicSheet.ts, skillSheet.ts, and spiritMagicSheet.ts

### DIFF
--- a/src/items/RqgItemSheet.ts
+++ b/src/items/RqgItemSheet.ts
@@ -1,13 +1,12 @@
-import { getGame, getRequiredDomDataset, localize } from "../system/util";
+import { ItemTypeEnum } from "../data-model/item-data/itemTypes";
+import { getGame, getRequiredDomDataset, localize, localizeItemType } from "../system/util";
 
 export class RqgItemSheet<
   Options extends ItemSheet.Options,
   Data extends object = ItemSheet.Data<Options>
 > extends ItemSheet<Options, Data> {
   get title(): string {
-    return `${localize("ITEM.Type" + this.object.type.titleCase())}: ${
-      this.object.name
-    }`;
+    return `${localizeItemType(this.object.type)}: ${this.object.name}`;
   }
 
   public activateListeners(html: JQuery): void {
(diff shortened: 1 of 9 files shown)


src/items/RqgItemSheet.ts
```diff
@@ -70,4 +73,46 @@ export class RqgItemSheet<
...
+    super._onDrop(event);
+    // Try to extract the data
+
+    if (
+      this.item.data.type === ItemTypeEnum.Cult ||
+      this.item.data.type === ItemTypeEnum.Rune ||
+      this.item.data.type === ItemTypeEnum.RuneMagic ||
+      this.item.data.type === ItemTypeEnum.Skill ||
+      this.item.data.type === ItemTypeEnum.SpiritMagic
+    ) {
```
wake42, line 87: I'd rather not have ifs like that if possible. I guess the problem is that we want to extend for example `CultSheet` with both `RqgItemSheet` and some class with that `_onDrop` function (and you can only extend from one place). I see two possible ways - 
Either make a `RqgItemSheetDescribed` class that extends `RqgItemSheet` and adds the `_onDrop`. This is a bit clunky though in case we ever wanted to add something else.
Or explore mixins. I havn't used them, but I know Foundry itself uses [mixins for Documents](https://foundryvtt.com/api/ClientDocumentMixin.html). With a `hasDescription` mixin that implements the `_onDrop` function one could add that to the sheets that should have it.

This is typescripts documentation about mixins https://www.typescriptlang.org/docs/handbook/mixins.html

I did a quick search and found the starfinder system also uses mixins
https://github.com/wildj79/foundryvtt-starfinder/blob/development/src/module/actor/mixins/actor-conditions.js
https://github.com/wildj79/foundryvtt-starfinder/blob/6191ee7be6072d6e62fad8bd4207b6faba4cad38/src/module/utils/custom-mixer.js#L1
https://github.com/wildj79/foundryvtt-starfinder/blob/6191ee7be6072d6e62fad8bd4207b6faba4cad38/src/module/actor/actor.js#L20

Do you think this could be something to try out?

wake42 (commented):
In addition I realised that maybe we should have a way of removing the description as well. In the weapon -> skill link there is a little trashcan icon after the link. Maybe we should have that here as well.

wake42 (approved):
I'll approve this and continue looking for a working typescript mixin pattern later.

Issue #170, opened 2022-01-21
Background "tab" can be pushed beyond the scroll


Moonpile:
![image](https://user-images.githubusercontent.com/3959760/150603033-ee58bf2d-632b-4b3f-b654-346be44f1ce6.png)

If you (somehow) have Spirit Magic, Rune Magic, and Sorcery, then you have to scroll to see "Background".

Make less space between the sidebar tabs.

Make the default character sheet size configurable?

wake42:
I think this might be fixed already. Need to verify though

Moonpile:
I no longer think this is a real issue.  Yes, if you have ALL THE TABS then Background hangs down, but what else is it supposed to do?  Also you'd probably have more notes on a character like that and it wouldn't look weird.  I recommend we just close this.

Moonpile:
Closed as invalid

PR #187, opened 2022-02-12
Issue77-drag-items-between-actors

Moonpile:
Handles _onDropItem in RqgActorSheet and allows for copying intangible items or transfering (giving) physical items (ie those with quantity).  If the quantity is greater than one allows transferring an arbitrary amount, "splitting the stack".

Note that a player cannot transfer items from an Actor Sheet they own to one that they do not, even if they have "limited" or "observer" permissions on it.  I've asked about this on the Foundry Discord, but no answer as of now.  The workaround may be for the party to have a "Party Loot Sheet" that they all have "owner" permissions to and copy items to and from there to exchange them.

 If the ability to drag intangible items (like runes or passions) is annoying to some we could make a setting that governs whether or not to express the "item-list" on the top level list element to enable or disable that behavior.

### DIFF
--- a/src/actors/item-specific/runeMagic.ts
+++ b/src/actors/item-specific/runeMagic.ts
@@ -23,9 +23,14 @@ export class RuneMagic extends AbstractEmbeddedItem {
     if (item.data.data.cultId) {
       const runeMagicCult = actor.items.get(item.data.data.cultId);
       if (!runeMagicCult || runeMagicCult.data.type !== ItemTypeEnum.Cult) {
-        const msg = localize("RQG.Item.Notification.ActorDoesNotHaveCultOnRuneMagicWarning");
-        ui.notifications?.warn(msg);
-        console.warn(msg, item, actor);
+        // This warning can happen when drag-dropping a rune spell from one Actor to another,
+        // but the notification happens a lot of times and doesn't really matter since the system immediately,
+        // displays the "Which cult provides this Rune Magic?" dialog allowing the player to fix it.
+
+        // const msg = localize("RQG.Item.Notification.ActorDoesNotHaveCultOnRuneMagicWarning");
+        // ui.notifications?.warn(msg);
+        // console.warn(msg, item, actor);
+
         item.data.data.cultId = ""; // remove the mismatched link to make it appear in the GUI
       }
       if (runeMagicCult && runeMagicCult.data.type === ItemTypeEnum.Cult) {
@@ -49,7 +54,9 @@ export class RuneMagic extends AbstractEmbeddedItem {
         (i) =>
           i.type === ItemTypeEnum.Rune &&
           (runeMagicRuneNames.includes(i.name ?? "") ||
-            (runeMagicRuneNames.includes(getGame().settings.get("rqg", "magicRuneName") as string) &&
+            (runeMagicRuneNames.includes(
+              getGame().settings.get("rqg", "magicRuneName") as string
+            ) &&
               cultRuneNames.includes(i.name ?? "")))
       )
       // @ts-ignore r is a runeItem TODO rewrite as reduce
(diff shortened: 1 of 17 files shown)


wake42:
For discussion:
When dragging an item from the sidebar to an actor - should it be removed from the sidebar then? Maybe this should follow the same rules as when dragging between actors so that especially "unique" items don't get duplicated. I guess this comes down to how one views what the items in the sidebar are. Either they are unique items that exist in the world (but is not owned by anyone yet) and therefore can't exist both in the sidebar and embedded in an actor.
The other view would be that the sidebar is not part of the world - it is just a technical storage of items of different types. If one views it this way, then the current behaviour with copying items to actors is correct.
What do you think?

Moonpile:
> 

I think the Foundry standard is that items on the sidebar are effectively "item templates".  If we were going to have special behavior I think there should be an actual "unique" flag beyond just there being a quantity of 1 and that should be reserved for things that really are "unique", like named magical items.  Just an ordinary Broadsword should stay on the sidebar when you drag it to a character.

Moonpile:
> The context menus don't work anymore

If that's the only requested change then I'd like to suggest that we close and merge this and deal with that problem under the issue #170, which I consider to include any issues related to the context menu as well as the scrolling issue.

wake42:
I had a look and found an easy fix for the broken context menus. Just remove `overflow-y: auto;` here https://github.com/sun-dragon-cult/fvtt-system-rqg/blob/52e430a589626b16582553fade52bb372d13176a/src/rqg.scss#L647 i didn't see any negative effects from that. (and I don't think #170 has anything to do with this?)

About the unique sidebar items - I agree that foundry standard is that it's just a template. And If we implement a "container" actor/sheet then it would be possible to drag from there instead, so let's keep it as it is.  

If you drag a weapon to an actor you should get the skill as well (unless the receiving actor don't already have it). Maybe can reuse the code from embedding a weapon https://github.com/sun-dragon-cult/fvtt-system-rqg/blob/52e430a589626b16582553fade52bb372d13176a/src/items/weapon-item/weaponSheet.ts#L258

Moonpile:
> I had a look and found an easy fix for the broken context menus. Just remove `overflow-y: auto;` here
> 
> https://github.com/sun-dragon-cult/fvtt-system-rqg/blob/52e430a589626b16582553fade52bb372d13176a/src/rqg.scss#L647
> 
> i didn't see any negative effects from that. (and I don't think #170 has anything to do with this?)

Ok, that at least puts it back to where we were with the context menu possibly wrapping in the column.  The reason I said it was related to #170 is because I was thinking of #170 as a larger issue about the context menus as well as the background menu being pushed beyond the scroll, or rather I thought those two were related, but I guess they're not.

Moonpile:
> If you drag a weapon to an actor you should get the skill as well (unless the receiving actor don't already have it). Maybe can reuse the code from embedding a weapon

Weapon.onEmbedItem was already firing and calling embedLinkedSkill, but when dragging from actor to actor the embeddedSkillId actually has a value so I just needed to remove the check to ensure that it DIDN'T have a value.

src/actors/item-specific/weapon.ts
```diff
@@ -82,14 +82,27 @@ export class Weapon extends AbstractEmbeddedItem {
     skillOrigin: string, // Linked skill item origin (uuid)
     actor: RqgActor // The actor that should have the skill
   ): Promise<string> {
-    if (!embeddedSkillId && skillOrigin) {
+    // If the weapon has the aspect (ie one hand, or missile):
+    // embeddedSkillId will be null when dragging from sidebar or directly from compendium
+    // embeddedSkillId will have a value when dragged from one actor to another
+    if (skillOrigin) {
```
wake42, line 88: I think this might give some problems in the case where the skillOrigin doesn't exist anymore? In that case you will get a notification - but there is no way a GM can fix that I think. We can try this out though.

wake42 (approved):

PR #189, opened 2022-02-14
Refactor journal entry onDrop

wake42:
I didn't manage to implement mixins, so I did it this way instead. It this an ok architecture? The intent is that you should only have to look at an item sheet to see what affect it.


Moonpile (approved):
Looks good to me.

PR #190, opened 2022-02-14
Moonpile/feat-Dont-auto-import-calendar-#188

Moonpile:
Removed the module support directory and stopped the system from trying to re-import a calender every time it reloads.  

Calendar support will be moved to a separate module, TBD.


Moonpile:
[RQG_calendar_support.zip](https://github.com/sun-dragon-cult/fvtt-system-rqg/files/8064719/RQG_calendar_support.zip)

The above are the calendar import files and module support I was working on when we decided to move them to a different module. Putting them here so they don't just get lost.

wake42 (approved):

PR #192, opened 2022-02-17
bug Rune Spell with rune the actor does not have causes an error #191

### LINKED ISSUE #191: bug Rune Spell with rune the actor does not have causes an error
For example "Makarios, son of Tharassa" has Resurrect which has the Fate rune which he does not possess.  Attempting to cast that spell clicking or double clicking (with or without the card) or using the context menu options for the same things causes the following error:

image

Moonpile:
fixes #191

### DIFF
--- a/src/chat/runeMagicCard.ts
+++ b/src/chat/runeMagicCard.ts
@@ -75,8 +75,10 @@ export class RuneMagicCard {
     // Get the actor's versions of the runes, which will have their "chance"
     usableRuneNames.forEach((rune) => {
       const actorRune = actor.items.getName(rune);
-      assertItemType(actorRune?.data.type, ItemTypeEnum.Rune);
-      runesForCasting.push(actorRune.data);
+      if (actorRune !== undefined) {
+        assertItemType(actorRune?.data.type, ItemTypeEnum.Rune);
+        runesForCasting.push(actorRune.data);
+      }
     });
 
     const strongestRune = runesForCasting.reduce(function (prev, current) {
@@ -118,9 +120,9 @@ export class RuneMagicCard {
         skillAugmentationOptions: skillAugmentationOptions,
       },
     };
-    
+
     await ChatMessage.create(await this.renderContent(flags));
-    activateChatTab()
+    activateChatTab();
   }
 
   public static async inputChangeHandler(ev: Event, messageId: string): Promise<void> {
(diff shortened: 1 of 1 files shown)


src/chat/runeMagicCard.ts
```diff
@@ -75,8 +75,10 @@ export class RuneMagicCard {
     // Get the actor's versions of the runes, which will have their "chance"
     usableRuneNames.forEach((rune) => {
       const actorRune = actor.items.getName(rune);
-      assertItemType(actorRune?.data.type, ItemTypeEnum.Rune);
-      runesForCasting.push(actorRune.data);
+      if (actorRune !== undefined) {
+        assertItemType(actorRune?.data.type, ItemTypeEnum.Rune);
+        runesForCasting.push(actorRune.data);
+      }
```
wake42, line 81: I think this is something that should never happen since you should get the runes of your cult when you drag the cult to the actor. Also you should only be allowed to learn rune magic that you can cast. So minimum I would suggest a ui notification warn that something is wrong. Maybe should investigate how it's possible to get into the state of casting rune magic with runes you don't have and see if it's possible to prevent that "back door".

wake42 (changes requested):

src/chat/runeMagicCard.ts
Moonpile: The Rune Magic Spell can have runes associated with it that the Actor who has the spell does not have.  I think that's a normal, albeit rare, situation.  In the case of Resurrection it has the Fate rune which starting characters don't normally get.

In the case of Makarios and Resurrection, since he does not have Fate but does have Harmony and Man, I thought about trying to pass the Fate rune to the list but somehow have it "disabled" maybe with a tool tip explaining why, but at that point I only have the name of the Rune as a string.  So if we wanted to get the Fate rune definition from there, how would we know where to look?  Can we assume the Runes compendium is there?  It's from the Starter set.

As it is right now (after my commit) Makarios cannot cast Resurrection using the Fate rune, but can cast with either Harmony or Man.  Before my fix you can't get the Resurrection card to display at all because of an error.

Maybe the thing to do is have an element on the Rune Magic Card that says "This character does not have one or more of the runes associated with this spell" and only show it in this situation?

wake42: I see that Resurrect on Makarios starter set sheet, and in book 1 has the fate rune - but resurrect in RBM does not, is this just a typo? 
RMB:
image

Book1:
image

Moonpile: Oh, Chaosium!

Anyway, I think it's an edge case but entirely possible that you'd have a rune spell with multiple runes and you wouldn't have one of them.

wake42 (approved):
Lets leave it like this - typo or no typo :-)

PR #209, opened 2022-03-11
fix: Added Punch key

lozanoje:


wake42 (changes requested):
Could you please add "Kick" as well.

lozanoje:
Done

wake42 (approved):

Issue #199, opened 2022-03-05
System specific item id


Moonpile:
As developers, we need a way to be able to reference items in compendia regardless of the localization. So the Starter Set in English might have a "Broadsword" and in Spanish a "Sable", but the Character Creation Wizard needs to be able to add it to the new Actor  based on an Occupation that has an item with "rqid" of "broadsword-starter" in its list of starting gear.

There should be a util function for finding items in the installed compendia based on the "rqid" provided.  This should error when none are found.  If one is found, it should return it. If more than one is found it might offer a dialog to allow the user to pick one before returning it.

Moonpile:
Some other thoughts:

- The GM should be able to see and edit the rqid, but it should be in a sort of "out of the way" place on the item sheet like a "system" tab
- (maybe) When a player's Actor owns an item they should also be able to edit the rqid like the GM can so that they can "keep their items separate", however the system shouldn't be looking on Actors for these items anyway.

wake42:
I think that related to the `rqgId` field there should also be a `language` and a `prio` field. With these 3 fields on all items it would be possible to find the correct item from compendiums by returning the item with the correct rqgId in a specific language (possibly with some fallback language like "en") that has the highest prio.

The prio field is needed to make for example the items in a RBM module to take precedence over starter set items if the same item (same rqgId) exists in both.

The default compendium index should include these 3 fields as well to make it possible to quickly find the correct item without having to search through all compendiums.

Moonpile:
> there should also be a `language` and a `prio` field.

The language one makes sense, I suppose. That way if they have, say, English and Spanish starter set installed, but the world is set to Spanish it will retrieve Spanish language items, though the GM can look through the English language compendia for things (and also the eventual compendium browser will filter on that).

So I guess with `prio` your idea is that the Starter Set might have a "broadsword" with a `prio` of, say 10, and the Core set might have a "broadsword" with a `prio` of 20, and so it returns the one with the higher priority.  And later if we do a "Weapons and Equipment" mod we might decide, well they tweaked the Broadsword and this is now definitive, so we give it a `prio` of 30 and if you have all three mods installed and look for a "broadsword" it will return the one from the "Weapons and Equipment" mod.   

We still have the issue of what if they're equal but at some point you just have to pick the first one from the result set, and also the priority shouldn't be equal.

Ok, if my re-statement of this agrees with what you were intending to say then I think we're on the same page.

wake42:
[...]

PR #216, opened 2022-03-16
getItemByRqid follow up

### LINKED ISSUE #215: getItemByRqid follow up
> Actually I realised one more thing. If you search for an empty "" rqid you will get a warning that there are more than one result, and you will get one (random) result. Maybe we should just return nothing if rqid is empty.

> Also I think I found a minor bug:
> 
> ```
> > await CONFIG.Item.documentClass.getItemByRqid("rune-darkness-element","en")
> rqgItem.js:159 Item not found in world or Compendia matching the rqid "rune-darkness-element", and rqidLang "en". See console log.
> rqgItem.js:169 RqgItem {apps: {…}, _sheet: null, parent: null, pack: 'starter-set-en-rqg.runes', data: ItemData}
> RqgItem {apps: {…}, _sheet: null, parent: null, pack: 'starter-set-en-rqg.runes', data: ItemData}
> ```
> 
> It displays an error but still finds the item in the compendium.

Moonpile:
Fixes #215
- if rqid parameter is undefined or empty return undefined
- Don't warn about not finding an item in getItemFromWorldByRqid.  Only warn if nothing is found after checking the world and the comendia
- removed extraneous console.log

fixed issues found after approving merge to #199

### DIFF
--- a/src/items/rqgItem.ts
+++ b/src/items/rqgItem.ts
@@ -174,7 +174,7 @@ export class RqgItem extends Item {
     rqid: string,
     lang: string = "en"
   ): Promise<RqgItem | undefined> {
-    if (!rqid || rqid === "") {
+    if (!rqid) {
       return undefined;
     }
 
--- a/src/items/rqgItem.ts
+++ b/src/items/rqgItem.ts
@@ -174,7 +174,6 @@ export class RqgItem extends Item {
     rqid: string,
     lang: string = "en"
   ): Promise<RqgItem | undefined> {
-
     if (!rqid || rqid === "") {
       return undefined;
     }
@@ -197,14 +196,12 @@ export class RqgItem extends Item {
       ui.notifications?.warn(msg);
       console.log(msg);
     }
-
   }
 
   private static async getItemFromWorldByRqid(
     rqid: string,
     lang: string = "en"
   ): Promise<RqgItem | undefined> {
-
     if (!rqid) {
       return undefined;
     }
@@ -221,9 +218,11 @@ export class RqgItem extends Item {
       let result = candidates.reduce((max, obj) =>
         max.data.data.rqidPriority > obj.data.data.rqidPriority ? max : obj
       );
-      
+
       // Detect more than one item that could be the match
-      let duplicates = candidates.filter(i => i.data.data.rqidPriority === result.data.data.rqidPriority);
+      let duplicates = candidates.filter(
+        (i) => i.data.data.rqidPriority === result.data.data.rqidPriority
+      );
       if (duplicates.length > 1) {
         const msg = localize("RQG.Item.RqgItem.Error.MoreThanOneRqidMatchInWorld", {
           rqid: rqid,
@@ -243,11 +242,10 @@ export class RqgItem extends Item {
     rqid: string,
     lang: string = "en"
   ): Promise<RqgItem | undefined> {
-    
     if (!rqid) {
       return undefined;
     }
-    
+
     const candidates: RqgItem[] = [];
 
     for (const pack of getGame().packs) {
@@ -262,7 +260,6 @@ export class RqgItem extends Item {
     if (candidates.length === 0) {
       return undefined;
     }
-
 
     if (candidates.length > 0) {
       let result = candidates.reduce((max, obj) =>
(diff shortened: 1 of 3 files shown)


src/items/rqgItem.ts
```diff
@@ -174,6 +174,11 @@ export class RqgItem extends Item {
     rqid: string,
     lang: string = "en"
   ): Promise<RqgItem | undefined> {
+
+    if (!rqid || rqid === "") {
```
wake42, line 178: The second part is unnecessary since empty string is "falsy"
```suggestion
    if (!rqid) {
```

src/items/rqgItem.ts
```diff
@@ -182,13 +187,28 @@ export class RqgItem extends Item {
...
+    }
+
   }
 
   private static async getItemFromWorldByRqid(
     rqid: string,
     lang: string = "en"
   ): Promise<RqgItem | undefined> {
+
+    if (!rqid || rqid === "") {
```
wake42, line 208: The second part is unnecessary since empty string is "falsy"
```suggestion
    if (!rqid) {
```

src/items/rqgItem.ts
```diff
@@ -229,13 +243,17 @@ export class RqgItem extends Item {
     rqid: string,
     lang: string = "en"
   ): Promise<RqgItem | undefined> {
+    
+    if (!rqid || rqid === "") {
```
wake42, line 247:
```suggestion
    if (!rqid) {
```

wake42 (changes requested):

wake42:
I've started entering the rqids in the starter set and I found one more issues:
for skill items - remove category and add specialisation so `skill-ride-horse` (instead of `skill-ride-agility`)

Moonpile:
- Hit Locations will most likely need to be edited after generating rqid to add "body form" (ie humanoid or quadraped or Troll)
- Runes include the rune type, for example "rune-beast-form"
- Cult is not a special case, for example "cult-lhankor-mhy"
- Rune Magic is not a special case but handles specialization, for example "rune-magic-command-cult-spirit", or without specialization "rune-magic-multispell"
- Spirit Magic is the same as Rune Magic: "spirit-magic-detect-substance" or "spirit-magic-fanaticism"
- Skills never show the skill group, just the name and if it exists the specializaion: "skill-drive-chariot" or "skill-jump"
- Gear is not a special case: "gear-lunar". **Should it be?  Should it include "Unique", "Consumable", or "Currency" item type?**
- Weapon is not a special case: "weapon-battle-axe"
- I thought armor should include the name prefix if it's there so we can put "special" items in the modules more easily so if the prefix is "horned" then "armor-horned-full-helm-bronze-plate", and if it's blank then "armor-full-helm-bronze-plate"
- Passions are not a special case: "passion-hate-lunar-empire" or "passion-honor"

Moonpile:
> * **Should it be?  Should it include "Unique", "Consumable", or "Currency" item type?**

I didn't think this added significant value to the rqid so I didn't do this.

src/system/util.ts
```diff
@@ -349,3 +352,29 @@ export function activateChatTab() {
   // @ts-ignore 0.8 tabs
   ui?.sidebar?.tabs.chat && ui.sidebar?.activateTab(ui.sidebar.tabs.chat.tabName);
 }
+
+
+export function getDefaultRqid(Item: RqgItem): string {
+  if (Item.type === ItemTypeEnum.Skill) {
```
wake42, line 358: I think if you check
```suggestion
  if (Item.data.type === ItemTypeEnum.Skill) {
```
then you don't need to do the assertItemType to get to make item narrowed to a skill item. Same in the other if:s

src/system/util.ts
```diff
@@ -349,3 +352,29 @@ export function activateChatTab() {
   // @ts-ignore 0.8 tabs
   ui?.sidebar?.tabs.chat && ui.sidebar?.activateTab(ui.sidebar.tabs.chat.tabName);
 }
+
+
+export function getDefaultRqid(Item: RqgItem): string {
```
wake42, line 357: `item` should be lowercase

wake42 (changes requested):
Regarding gear I agree that the type (unique etc) should not be part of the rqid.

src/system/util.ts
Moonpile: done

src/system/util.ts
Moonpile: done

wake42 (approved):

PR #217, opened 2022-03-19
Create Homeland Item

### LINKED ISSUE #202: Create Homeland Item
To make creating an extensible character wizard #101 possible we need a way of storing different homelands and their impact on characters. Adding a homeland item would make it easy to include a set of homelands in a compendium

The homeland item should contain
- Description (large text) _- should this be a link to a journalEntry to keep the data size embedded in an actor down?_
- Cultural Stereotype (large text) _- link to journalEntry?_
- Common Cults (list of references to "Cult" items by rqgId #199)
- Suggested Occupations (list of references to "Occupation" items by rqgId #199)
- Passions (list of references to "Passion" items by rqgId #199) 
- List of presets for Town / Home Area dropdown datalist
- Characteristics modifiers (SIZ / DEX +-)
- Map of cultural skill bonuses (rqgId -> modifier). _For weapon skills this should be checked when learning (adding) a new weapon skill._

If we're making a "Homeland" item type do we just have a field on the Actor to hold a Homeland? Then just render it on the background tab, replacing some existing fields there. Also need to give a way of copying a cult over to the Actor, and copying over the base passions.
 
### Additional context
How should we handle dragging a homeland to an actor that already has a homeland item?

### Related to
- [ ] #203

### Blocks
- [ ] #101

### Blocked by
- [x] #199

Moonpile:
Fixes #202

- added Homeland Item type


Moonpile:
On the homelandSheet.ts, I copied the activateListeners exactly from runeSheet.ts.  Should the drop and journal link code be moved to RqgItemSheet so it's available for all item sheets?

Why is "showJournalEntry" a method of RqgActorSheet?  I'm starting to think we need a class and hbs for journal and rqid links so that they're entirely modular.

It feels like there should be a generic way to handle taking the drop target, which is the name of a property, and using it to persist the journal.  Unfortunately I couldn't figure out how to use a variable or interpolated strings to create the update data records.  If you have any ideas how to avoid the if statemetns there and make it generic to save a journal item to a field when we know the field name from the target, that would be helpful since I think we'll be doing this again.

I had to move the getItemByRqid from RqgItem to utils because having it there was causing the problem with the passion sheet initializing (and I suspect passion was just the first one we hit).

Areas for improvement:

- It would be cool if the rqgidLinks showed the icon of the item (or at least the default icon of the item) instead of the suitcase. It might be too small though.
- Journal rqid
- Make journal and rqid links into a more modular drop-in element
- We need a compendium for the Passion items.
- We need a lot of journal entries for homelands, regions, tribes, clans, cultures

src/data-model/shared/journalentrylink.ts
```diff
@@ -3,4 +3,5 @@ export class JournalEntryLink {
   journalId: string = "";
   /** compendium pack (if it's in a compendium) */
   journalPack: string = "";
+  journalName: string = "";
```
wake42, line 6: It would be better to not use JournalEntryLink. That was a mistake from my side since I didn't know about uuid at that time. If you have a uuid there is also a function in `util.uuid2Name` that gives the name and some extra text (that might need a little refactoring). That way there is no need to store the name of a linked journalEntry.

wake42 (changes requested):
It would be much better to use uuid instead of the JournalEntryLink interface. 

As it is now each homeland item for homeland + city would be a separate item. I though the idea was that you would have links there so you (later on the actor wizard sheet) could select which city you wanted depending on what homeland you chose. This way you would be presented with a very long list of homeland items.

Just for noting it down - can be fixed later:
* name of homeland item is not updated immediately when dragging JE to it.
* There should be a way of deleting added journal entry links

src/data-model/shared/journalentrylink.ts
Moonpile: > use uuid instead of the JournalEntryLink interface.

So we would have to create our own handler to look for the data-uuid attribute?  Or does the built-in journal link functionality handle that somehow?  

> homeland + city

If we make the homeland just be "Sartar" then we could do a list of links to regions the way it does tribes and clans right now, but then where do we put the tribes and clans and the rest of the stuff?  All of the fields in the Homeland tab actually pertain to the "Region or City" and (with the exception of Cults which is some from Homeland and some from Region).  Maybe the Homeland tab should really have been called "Region" the way I have it now.

We could have a "Region" item or structure that is a child of homeland and has a link to the journal and all of the fields on Homeland the Homeland tab and they get repeated for each "Region".  Sartar has 12 Regions, so I think the form for that would be unwieldy.  Also, if the Actor sheet is going to hang on to the Homeland object after character creation so that the player has a handy reference to things they might know, I think it makes sense for someone from Colymar to know a lot of detailed information about those lands but if they have links to things in Alone and Wilmskirk and so on, it's both too much information for the player and probably might be more than their character would reasonably know.

In the Wizard I was envisioning that we'd look through the journals (in both the world and all compendia?) for Homeland objects and group them by the data.homeland values, so we could present a list like "Grazelands, Lunar Tarsh, Prax, Sartar, etc".  The player picks one of those, and then we look for journals with that data.homeland and present a list based on the data.region, like "Alda-Chur, Alone, Boldhome, Colymar Tribe, etc".  

I agree that it creates a lot of Homeland objects, 12 for Sartar, but I think it keeps them in more manageable chunks of information, especially since the bulk of the really important stuff derives from the Region.  Everything about the overall Homeland (like Sartar) is pretty much free-form text that can be put into a shared journal article.

Moonpile:
> * name of homeland item is not updated immediately when dragging JE to it.

Fixed

Moonpile:
> * There should be a way of deleting added journal entry links

I think we should try to move forward with this so we can get to the creation wizard.  Anything we do for this would be a global refactor of all journal links to be able to delete them from where ever they are.

src/data-model/shared/journalentrylink.ts
wake42: > So we would have to create our own handler to look for the data-uuid attribute? Or does the built-in journal link functionality handle that somehow?

When an item is dropped, the uuid to to that item is included in the provided data. If you have the uuid you can lookup the name (see util.uuid2Name() ) so there is no need to store that.  And for generating the html a uuidLink.hbs would do it, there is some functionality built in to Foundry in [enrichHTML](https://foundryvtt.com/api/TextEditor.html#.enrichHTML) but that is more for parsing journal style links like `@Compendium[rqg.demo-actors.ro196XqFnrNvYB69]{Demo Spirit}` to html links with icon.

The weapon item already use uuid. That said I agree that this can be looked at later (even though it hurts a little to add more fields to all items that use the journalLink interface already. The important thing is to get to the point where this can be tested in a simple character creator.

> I agree that it creates a lot of Homeland objects, 12 for Sartar, but I think it keeps them in more manageable chunks of information, especially since the bulk of the really important stuff derives from the Region. Everything 

Yes after your explanation I agree with you. The character creation wizard would include a specialised version of a compendium browser (using getItemByRqid). First display a list of Homelands (derived from the list of all homeland items) and then after one of them is selected, show either the actual homeland items or a dropdown with regions.
I think we need one more "find rqid" function that returns a list so that one can get the homelands to choose from (but that's in another case...).

wake42 (approved):
We'll come back to this later, but I'm approving for now so that we can get to the point of testing this.

Moonpile:
> We'll come back to this later, but I'm approving for now so that we can get to the point of testing this.

I will try to work out the "link by uuid"  process while working on the Occupation and if we're happy with that we can back-port it into the Homeland item (and all the other journal links).

PR #224, opened 2022-03-22
Create Occupation Item

### LINKED ISSUE #203: Create Occupation Item
To make creating an extensible character wizard #101 possible we need a way of storing different occupations and their impact on characters. Adding an occupation item would make it easy to include a set of homelands in one or more compendiums.

The Occupation item should contain
- Description (large text) link to journalEntry
- ~~Homelands (list of "Homeland" items by ID? or just text?~~
- Occupational Skills (map of "Skill" items by Compendium ID and bonus as integer and whether or not the skill is used for income)
- Standard of Living (text, fills in field on background tab)
- Base Income (text, fills in field on background tab)
- Cults (list of "Cult" items by ID? Or just text?)
- Favored Passions (list of "Passion" items by ID?) (player chooses one at +60% or increases one they already have by +10%)
- Ransom (text, fills in field on background tab)
- Equipment (list of "Armor", "Gear", and "Weapon" items by ID)

### Additional context
If we're making an "Occupation" item type, do we just have a slot on the Actor for an array of Occupations and render them that way? On the actor we would only need to render the description, standard of living, base income, and ransom. During creation there needs to be some way to add the skills (or update existing skills), add a cult (optional as this could have been done under Homeland already), add or update a favored passion, and copy all the equipment over to the Actor.

### Related to
- [x] #202

### Blocks
- [ ] #101

### Blocked by
- [x] #199

Moonpile:
Fixes #203

- Created Occupation item type and related sheets and classes
- renamed "data-journal-target" to "data-target-drop-property" since it handles items and journal entries beging dropped to item properties.


Moonpile:
I pulled out the TODO about moving StandardOfLivingEnum to its own file.  If you want me to do that, let me know where you think it belongs.

src/data-model/item-data/occupationData.ts
```diff
@@ -0,0 +1,63 @@
...
+
+export enum StandardOfLivingEnum{
+  Destitute = "destitute",
+  Poor = "poor",
+  Free = "free",
+  Noble = "noble",
+  PettyKing = "petty-king",
+}
+
+export class occupationalSkill {
```
wake42, line 15:
```suggestion
export class OccupationalSkill {
```

src/data-model/item-data/occupationData.ts
```diff
@@ -0,0 +1,63 @@
...
+  data: OccupationDataSourceData;
+}
+
+export const emptyOccupation: OccupationDataSourceData = {
+  occupation: "",
+  occupationJournalLink: new JournalEntryLink(),
+  specialization: "",
+  homelands: [],
+  occupationalSkills: [],
+  standardOfLiving: StandardOfLivingEnum.Destitute,
```
wake42, line 54: Maybe set the default to `Free` instead since that is probably the most common value

wake42 (approved):
I made 2 super minor comment - no need to reapprove for that after you've fixed them so I set this as approved now. There are some issues, mostly UX oriented I think, but this contains what is needed for continuing on the wizard path. We will have to come back to these eventually though.

I didn't get the new default icons in my "old" test world, and I see why. We are only looping through the setting values in the default icon setting application - not the available items. I will open a separate case about that.

Issue #101, opened 2021-12-20
Create Actor Wizard


wake42:
Implement some kind of wizard that can embed hitlocations / runes etc after answering some questions. Some inspiration could be taken from the [Chaosium wiki](https://rqwiki.chaosium.com/adventurers).

1. Choose race 
    - if human include human hitlocations
1. Player / NPC
    - player should get all skills, should a NPC only get a few?
    - add player runes
    - set reputation to 5%
1. Choose homeland
    - Add Passions
    - increase cultural skills
    - increase cultural runes

1. Choose Occupation
    - Add Passions
    - increase cultural skills
    - set background info like Standard of living etc

1. Characteristics
    - A step to choose which runes should be raised?

1. Cult
    - Show a list of cults in the world

1. Customise
    - Pick 4 skills and add +25% to each
    - Pick 5 skills and add 10% to each

Some additional thoughts. Players need to be able to run this wizard to make it possible for them to create their characters. Players don't have access to the "Create Actor" button in foundry so how should the players run this?
One idea is to add a state to the actor to show if a wizard has been run on it or not. The flow would then be that: 

* The GM creates an actor and gives the player access to it. 
* The player opens the actor sheet and since the wizard has not yet been run, it should display a big prominent "Initialize Character" button that runs the wizard.
* The GM should have another button / option "Skip Initialization" to make it possible to manually set the actor up.

...Other ideas are welcome :-)

Moonpile:
Capturing our discussion, one way to implement this might be to have a "Character Creation Sheet" (in addition to the normal Actor Sheet and eventual NPC and Loot sheets).  The sheet itself could function as a "tabbed wizard" using the same looking side-tabs that we use in the normal Actor Sheet, but the tabs would be the steps of character creation.  And when you're done you just change sheet to the normal Actor Sheet (or NPC).

Moonpile:
How would we store the data required for this so that it's expandable in the future?  For example, "Homeland" is fairly simple but it does add passions, skills, and runes.  I'm thinking it could be a json blob somewhere, but that raises two questions. First, is that extensible in the future?  Second, where do we put the blob?  In the starter set?  Or do we make a new item type of "Homeland" and have them in a compendium?  Then the Actor creation wizard can just search for all the "Item" compendium entries of type "Homeland".

We have the same basic problem with "Occupation" and whatever works for "Homeland" ought to work for "Occupation" as well.

[...]

PR #241, opened 2022-03-27
#101 Subtask: Step 0 Choose Species

### LINKED ISSUE #230: #101 Subtask: Step 0 Choose Species
On the alternate character sheet for Adventurer creation, allow the player to choose a species.  These will be defined as "template" actors in a specific compendium.

- Should there just be one canonical compendium where we put templates?
- Configuration allowing multiple compendia to search through?
- Maybe the actor can have an "isTemplate" property and we can look through all the compendia for an actor where isTemplate is true?

See #101

Moonpile:
Fixes #230

- added getActorTemplates to system.api
- RqgActorCreationShet extends ActorSheet directly rather than extending RqgActorSheet
- added creation-species.hbs
- added species-description.db compendium pack
- added species-templates.db compendium pack


Moonpile:
The commits I did on this were when we were trying to use the alternate actor sheet route.

PR #249, opened 2022-04-03
#101 Subtask: Actor Wizard FormApplication Framework

### LINKED ISSUE #246: #101 Subtask: Actor Wizard FormApplication Framework
Create the framework of the FormApplication for the Actor Creation Wizard.  This form will allow the player to make choices that affect an Actor they own while viewing the Actor in the standard Actor Sheet side-by-side so the player can see the effect of their choices.

- [ ] Should have some way of easily opening for the Player when the GM makes them a new Actor

See #101

Moonpile:
Fixes #246

- added actorWizardApplication.ts and actorWizardApplication.hbs
- added header button to actor sheet to launch actor wizard
- added sheet-button class
- added creation-sheet-tabs class
- added flagScope to config
- added actorWizardFlags to config
- added colonBreak handlebar helper

### DIFF
--- a/src/dialog/actorWizardApplication.hbs
+++ b/src/dialog/actorWizardApplication.hbs
@@ -11,26 +11,17 @@
 
   <div class="navbackground"></div>
   <nav class="creation-sheet-tabs tabs" data-group="primary">
-    <a class="item norse" data-group="primary" data-tab="0-species">
-      {{{colonBreak (localize "RQG.ActorCreation.Nav.0-Species")}}}</a>
-    <a class="item norse" data-group="primary" data-tab="1-homeland">
-      {{{colonBreak (localize "RQG.ActorCreation.Nav.1-Homeland")}}}</a>
-    <a class="item norse" data-group="primary" data-tab="2-family-history">
-      {{{colonBreak (localize "RQG.ActorCreation.Nav.2-FamilyHistory")}}}</a>
-    <a class="item norse" data-group="primary" data-tab="3-rune-affinities">
-      {{{colonBreak (localize "RQG.ActorCreation.Nav.3-RuneAffinities")}}}</a>
-    <a class="item norse" data-group="primary" data-tab="4-characteristics">
-      {{{colonBreak (localize "RQG.ActorCreation.Nav.4-Characteristics")}}}</a>
-    <a class="item norse" data-group="primary" data-tab="5-occupation">
-      {{{colonBreak (localize "RQG.ActorCreation.Nav.5-Occupation")}}}</a>
-    <a class="item norse" data-group="primary" data-tab="6-cult">
-      {{{colonBreak (localize "RQG.ActorCreation.Nav.6-Cult")}}}</a>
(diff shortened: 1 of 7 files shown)


src/dialog/actorWizardApplication.ts
```diff
@@ -0,0 +1,60 @@
...
+
+    this.form?.querySelectorAll("[data-actor-creation-complete]").forEach((el) => {
+      el.addEventListener("click", (ev) => {
+        this._setActorCreationComplete();
+      });
+    });
+  }
+
+  _setActorCreationComplete() {
+    this.actor.setFlag(RQG_CONFIG.flagScope, RQG_CONFIG.actorWizardFlags.actorWizardComplete, true);
```
wake42, line 51: The "Adventurer Creation" header button is not removed by this. It is no longer visible if I close and reopen the actor though. 
I tried forcing a rerender of the actorsheet by manually doing a actor.render(true) but that didn't remove the window header button either. Not sure how to fix this but it seems the window header is not reinitialised with just a rerender.

wake42 (changes requested):

src/dialog/actorWizardApplication.ts
wake42: I really need to get some sleep now, but I found this chat from Forien that relates to this https://discord.com/channels/170995199584108546/527196252576808960/729472401280204800 
Haven't looked into if / how he solved it...

Moonpile: Done

src/dialog/actorWizardApplication.hbs
```diff
@@ -0,0 +1,76 @@
...
+    <div class="topheaderline">
+      <h1>{{localize "RQG.ActorCreation.AdventurerCreationWizardTitle"}}</h1>
+      <h1 class="charname"><input name="name" type="text" value="{{actor.name}}" /></h1>
+    </div>
+  </header>
+
+  <div class="navbackground"></div>
+  <nav class="creation-sheet-tabs tabs" data-group="primary">
+    <a class="item norse" data-group="primary" data-tab="0-species">
+      {{{colonBreak (localize "RQG.ActorCreation.Nav.0-Species")}}}</a>
```
wake42, line 15: I think a better way to handle the newlines is to add a css class with `white-space: pre-line;` along with adding `\n` in the translation. That way it's more flexible where you want to put the newline. There is already a class `.nowrap` so maybe a `.preservewrap` could be added. This way the extra handlebar helper isn't needed.

wake42 (changes requested):
Sorry I didn't think of this one in my previous review...

src/dialog/actorWizardApplication.hbs
Moonpile: Done.  This does require that everything inside the <a> tag be on the same line, and I left a comment to that effect.  I'm concerned Prettier could break this on us.

wake42 (approved):

PR #251, opened 2022-04-05
Random Name Generator

### LINKED ISSUE #200: Random Name Generator
As a player or GM, I want to be able to create names for characters.  

This should be available in the character creation, but also as a way for the GM to create some names to chat window.

Ideas:
- use https://github.com/mrsharpoblunto/foswig.js/ for Markov chain generation of the actual name
- Use roll tables for epithets (like "The Red" or "Many-Horses")
- Markov input might come from journal entries in a compendium to make it easy to add, edit, and maintain name sets.

Moonpile:
Fixes #200

- added nameGeneration to api
- added dependency to Foswig
- referenced Rqid class in the system api for access to all methods
- rqidPrefixes defined in RQG_CONFIG
- getDefaultRqid uses rqidPrefixes
- rqidApi now supports getting JournalEntry and RollTable as well as Item


src/system/api/rqidApi.ts
```diff
@@ -1,126 +1,466 @@
...
 
-  const worldItem = await fromWorldRqid(rqid, lang);
+export class Rqid {
+  /**
+   * Return the highest priority Document matching the supplied rqid and lang from the Documents in the World. If not
+   * found return the highest priority Document matching the supplied rqid and lang from the installed Compendia.
+   *
+   * Example:
+   * ```
+   * game.system.api.fromRqid("someid", "es")
```
wake42, line 15: This is not correct anymore I think? Maybe we should just remove the example. Usually not a good idea to have code in comments anyway...

wake42 (changes requested):
I think the code example in the itemsheets are wrong as well?

src/system/api/rqidApi.ts
Moonpile: I removed the example and started a wiki page where we can document this: https://github.com/sun-dragon-cult/fvtt-system-rqg/wiki/rqid

And I had already corrected the item sheet.

wake42 (approved):

PR #257, opened 2022-04-17
#101 Subtask: Step 1 Choose Homeland

### LINKED ISSUE #231: #101 Subtask: Step 1 Choose Homeland
- Add a method for finding the all of the "best" items in the compendia by rqid and type.
- Use that to display homelands as a choice during character creation on the alternate character creation sheet.
- Allow the user a dialog to fine-tune what gets copied to the new Adventurer after choosing a homeland
- Copy the selected items to the Actor
-- Passions (see RQG p 27)
-- Cultural Skills (see RQG p 60)
-- Cultural Weapon Skills (see RQG p 60, this probably needs to be treated separately since later you get to pick bonuses to one cultural weapon in Occupation)
- Save Homeland choice to give suggestions for choosing Cult and Occupation later

Tangentially related:
- change Homeland "JournalEntryLink" properties to "RqidLink" properties.

See #101

Moonpile:
Fixes #231

- Reworking all JournalEntryLinks to RqidLinks
- RqidLink has wrappers for displaying an RqidLink property or an RqidLink array property and handling drag-drop and delete.

- WARNING: this is where I started to think something was wrong when all of the sudden "ConsolidateCompendiumItems" was not compiling.  I commented it out to build.


src/items/cult-item/cultSheet.ts
```diff
@@ -91,6 +91,5 @@ export class CultSheet extends RqgItemSheet<ItemSheet.Options, CultSheetData | I
 
   protected async _onDrop(event: DragEvent): Promise<void> {
     super._onDrop(event);
-    await droppableJournalDescription(this.item, event);
   }
```
wake42, line 94: This override is not necessary anymore I think.

src/items/rune-item/runeSheet.ts
```diff
@@ -99,6 +99,5 @@ export class RuneSheet extends RqgItemSheet<ItemSheet.Options, RuneSheetData | I
 
   protected async _onDrop(event: DragEvent): Promise<void> {
     super._onDrop(event);
-    await droppableJournalDescription(this.item, event);
   }
```
wake42, line 102: This override is not needed anymore.

src/rqg.scss
```diff
@@ -730,11 +730,11 @@ body {
...
-  .card-journal-link {
-    .invalid {
-      color: var(--rqg-color-invalid-journal);
-    }
-  }
+  // .card-journal-link {
+  //   .invalid {
+  //     color: var(--rqg-color-invalid-journal);
+  //   }
+  // }
```
wake42, line 737: Remove?

src/actors/rqgActorSheet.ts
```diff
@@ -1186,7 +1186,7 @@ export class RqgActorSheet extends ActorSheet<
         const link: RqidLink = {
           rqid: rqid,
           name: droppedJournal.name || "",
-          itemType: droppedEntityData.type,
+          documentType: droppedEntityData.type,
```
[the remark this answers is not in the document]
Moonpile: Fixed this everywhere.

src/items/cult-item/cultSheet.ts
Moonpile: Did you remove this and the other one below, because I don't have them anymore?  I also thought I got them all with the big commit.

src/rqg.scss
Moonpile: Removed

src/actors/context-menus/cult-context-menu.ts
```diff
@@ -1,37 +1,24 @@
...
     condition: (el: JQuery) => {
-      const itemId = getDomDataset(el, "item-id");
-
-      let firstItemEl = el[0];
-      while ((firstItemEl?.previousElementSibling as HTMLElement)?.dataset?.itemId === itemId) {
-        firstItemEl = firstItemEl.previousElementSibling as HTMLElement;
-      }
-      return !!firstItemEl.dataset.journalId;
+      const rqid = findDatasetValueInSelfOrAncestors(el[0] as HTMLElement, "cultRqid");
+      return rqid ? true: false;
```
wake42, line 15: I think you could simplify this to `return !!findDatasetValueInSelfOrAncestors(el[0] as HTMLElement, "cultRqid");`

Also in a bunch of more places

src/actors/context-menus/rune-context-menu.ts
```diff
@@ -70,22 +72,14 @@ export const runeMenuOptions = (
...
     icon: ContextMenuRunes.ViewDescription,
     condition: (el: JQuery) => {
-      const itemId = getRequiredDomDataset(el, "item-id");
-      let firstItemEl = el[0];
-      while ((firstItemEl?.previousElementSibling as HTMLElement)?.dataset?.itemId === itemId) {
-        firstItemEl = firstItemEl.previousElementSibling as HTMLElement;
-      }
-      return !!getDomDataset($(firstItemEl), "journal-id");
+      const rqid = findDatasetValueInSelfOrAncestors(el[0] as HTMLElement, "runeRqid");
+      return rqid ? true : false;
```
wake42, line 76: I think you could simplify this to `return !!findDatasetValueInSelfOrAncestors(el[0] as HTMLElement, "runeRqid");`

wake42 (changes requested):
Before you could click on the icon of items (like runes) to display the description and you got a ? cursor to hint about it. That doesn't work anymore (but the context menu "view description" works). The runes at the top of the actorsheet is still using the old description links (and actually still works though - complete with the ? cursor). I also think that that was a so and so thing UX wise. Maybe one could have a hover effect that displays a ⓘ on top of the item image or something like that top make it more discoverable that you can get the description that way.

I'm wondering if it would be possible to do some kind of migration to keep worlds using this now working. I guess to do that one would have to first include rqid in our compendiums and then go through all items in the world and embedded, to look up the description JE  with the old method, and then add the rqid of that to the flags of the item.

The advanced skills/runes section seem to have the wrong number of columns
image

src/actors/context-menus/cult-context-menu.ts
Moonpile: We will be messing with this again when/if we do the optional uuid based behavior.   Just so we're on the same page, if you drop a Document that doesn't have rqid stuff in the flags it could still link it (perhaps with a warning) and just save the uuid as a property of RqidLink.  In this case we'd have to put the uuid in the dataset if there's no rqid to put there, in which case this code will have to be re-worked to look for the uuid if it doesn't find an rqid.  So the question is, "Do we want this uuid backup behavior?  And if so should I do it now?"

src/actors/context-menus/rune-context-menu.ts
Moonpile: See above.

Moonpile:
> Before you could click on the icon of items (like runes) to display the description and you got a ? cursor to hint about it.

Yeah, I saw that but didn't clicking on the icon of the skill just bring up the description?  If we can do that via the context menu do we still want that?  The concern then becomes that the user might not know the context menu is there, but I'm not sure what to do about that.

Right now it says "Skill Description" as the title of the icon, but doesn't do anything, and also doesn't have a context menu.  So I agree it's not quite right.

What do we want?  Back the way it was?  Or make the icon clickable to roll, and right-clickable for context menu like the rest of the skill?

Moonpile:
> I'm wondering if it would be possible to do some kind of migration to keep worlds using this now working.

I think we should be able to do that, but we need to see what the "final form" of this is and then compare it with what we last released.  At the same time those people are beta-testing, so while it would be nice, it's not mandatory.  After we do a real release we'll have to do this kind of thing for changes like this.

Moonpile:
> The advanced skills/runes section seem to have the wrong number of columns

I hadn't even looked back at the wizard!  Certainly this will get fixed before I'm done.

Moonpile:
> The advanced skills/runes section seem to have the wrong number of columns

Committed fix for this.

src/actors/context-menus/cult-context-menu.ts
wake42: Yes I think a uuid "backup" is a good thing (needs some more thought though), and yes we could wait with that. Resolving this comment.

wake42:
About opening the description from the icon...

In my mind the context menu should contain all actions you can do with an item, but the primary actions should be accessible without opening that menu, and opening the description is a primary action in my mind. 

I also think the current solution of only changing the cursor to a ? is not a good enough affordance to make that behaviour discoverable. A bigger overlay with for example a ⓘ on top of the icon when you hover could be a better affordance, as an analogy there are systems that show a big dice on top where you can click to roll for example. Alternatively one could change the cursor to a url with a bigger image - that could maybe also be done for areas that are rollable (showing an image of a d100). Actually I think the cursor way is better - it just has to be more obvious than the small ? cursor.

I made a quick and dirty sketch to ty it out, there are 2 cursors in this picture hovering a description link and a roll link... 
image

Moonpile:
> I made a quick and dirty sketch to ty it out, there are 2 cursors in this picture hovering a description link and a roll link...

Ok, I like it.  Do we want just two actions?  Hover over the icon for "book" cursor with click meaning open the journal entry, hover over the name or chance for "dice" cursor with click meaning roll?  Hovering over the name could give an "edit" icon and a click could open the item.  The downside is it makes the area to click to roll smaller.

wake42:
> Hovering over the name could give an "edit" icon and a click could open the item.

We could try that out as well. As you say the roll area will become smaller, but on the combat table you can only click the percentage part anyway (since there are multiple per weapon), so in a way it will become more consistent. I think it's one of those things one would have to try out how it feels to know if it gets in the way or if it helps.

wake42 (changes requested):
Looks good! I've got some reflections:
- Not possible to drop rune from compendium on Homeland "rune links", same for skill links
- Rune and skill % increases are not reflected on actor runes & skills
image

- background tab clan field shows both clans and cults
image

- cult without rqid show up like lonely checkmark 
image

Moonpile:
> * Not possible to drop rune from compendium on Homeland "rune links", same for skill links

This works for me, but I think what might be happening is that if the Runes in your compendium do not have rqids it "sort of adds the rune".  

image

Notice that the circle i info thing is  gone, which indicates there's actually a rune on there.  

But then I added an rqid to the Air rune in my compendium and I was able to drag straight from the compendium to the Homeland.

image

So I think the thing to do here is guard against things being added to the Homeland that don't have rqids.

Moonpile:
> * Rune and skill % increases are not reflected on actor runes & skills
>   image

This works for me.  Is it possible the Air Rune on the homeland didn't have an rqg.rqid flag?

Moonpile:
> * background tab clan field shows both clans and cults
>   image

Fixed

Moonpile:
> * cult without rqid show up like lonely checkmark
>   image

After fixing the RqgItemSheet to not allow dropping items without rqg.rqid flags, this shouldn't be a problem anymore.

wake42 (approved):

PR #258, opened 2022-04-23
Implement building compendium packs from yaml files

wake42:
It is now possible to define the compendium packs as yaml files, located in `src/assets/compendiums`. Each compendium is represented by a folder that should be named as the resulting pack db file. The folder can contain one or more yaml-files and each yaml file can contain one or more documents.

The `yarn build` command now builds both the compendium packs and the system. To have a faster build you can use `yarn build:system` that only builds the system, but that depends on that you have built the packs once before.

The resulting pack db files are built to `src/assets/packs/` but they are gitignored since they shouldn't be comitted.


src/assets/compendiums/demo-actors.db/adventurer.yaml
```diff
@@ -0,0 +1,1672 @@
...
+    attribute: attributes.hitPoints.value
+  bar2:
+    attribute: attributes.magicPoints.value
+  randomImg: false
+items:
+  - _id: UQ8H7xAJIgpVG2Dk
+    name: Abdomen
+    type: hitLocation
+    img: systems/rqg/assets/images/items/hit-location.svg
+    data:
```
Moonpile, line 97: I believe the hit locations should have rqid stuff.  Given how we're discussing changing rqid to flags and also that this is more about setting up the build pipeline than it is the content, I'm ok with leaving this for now.  Is there some way we can define say "Abdomen" in one place and reference that on all actors and in the hit locations compendium?  I suppose they'd need unique _id values?

src/assets/compendiums/runes.db/runes.yaml
```diff
@@ -0,0 +1,618 @@
+name: Eternal Battle (condition)
+type: rune
+data:
+  rqid: rune-eternal-battle-condition
```
Moonpile, line 4: These rqids need "i." in front of them.  Again, this PR isn't really about the content so I'm ok with leaving it.  We definitely need to discuss the rqid naming schema.

src/assets/compendiums/name-bases.db/name-bases.yaml
```diff
@@ -0,0 +1,77 @@
+name: 'Names: Sartarite, Female'
```
Moonpile, line 1: Can we put flags onto these JournalEntry items here in the yaml?  That way we can flag the rqid.

`{
    "rqg": {
        "rqid": "je.name-bases-names-sartarite-female",
        "rqidPriority": 0,
        "rqidLang": "en"
    },
    "core": {
        "sourceId": "Compendium.rqg.name-bases.JNQLdEHBRqoXFcMx"
    }
}`

Moonpile (approved):
Given that this is laying the groundwork and not about the content of the compendia, this looks good to me.

The following come to mind as things we'll need to consider as we move toward refining and adding to the content.

- Can we define things in one place and reference them in many places? Like define "Abdomen" in just one place and reference it on all Actors (that need it) and in the Hit Locations compendium.  This would make sure that everyone has the same thing.  Problems recalculating hit point (in this case) and things would have the same _id values.
- Flags on JournalEntry items in YAML?
- rqid naming schema will be something we need to sort out soon.

PR #323, opened 2022-08-16
add cursors for roll (show chat) and open description

wake42:
I'm not sure about this change UX wise, so I'm just opening a draft PR te get some feedback.

I've added custom cursors for rolls and descriptions links, at the same time I've removed the "depressed" look when hovering since it's not possible to affect the whole row with the current grid layout.

Is the affordance too small and / or is the cursor images too much in the way? In short is this better or worse compared to what we currently have.

image
image


Moonpile:
Any chance we could wrap this in a feature flag? I can see people wanting to turn it off once their players know how it works.

TheChaosium:
It is good to see that clicking on the two items (the rune and the name) do different things, but if you are expecting to have any other actions then I would have left-click roll and right-click bring up a menu.

PR #340, opened 2022-08-24
339 create a system label file that will be common for all modules

TheChaosium:
This works with the system.json now being used.


wake42 (approved):
Sorry for committing in your branch - but it was easier to just update the error msg compared to explaining where that change should be...
With that change it's all good to go!

wake42:
A little note about merging this... this is not a `feat:` but a `refactor:` so remember to name the squash commit with that prefix.

PR #360, opened 2022-09-13
feat: Adding skill descriptions to the RQ System

TheChaosium:
Skill descriptions copied from the RQ starter Set module and moving in.
Changed references to always use the OpenSystem prio and lang

### DIFF
--- a/src/assets/pack-templates/skill-descriptions/agility.yaml
+++ b/src/assets/pack-templates/skill-descriptions/agility.yaml
@@ -1,0 +1,97 @@
+# boat
+_id: KRacu04VC87nhjOQ
+name: ${{openSystem.content.skill.boat.skillName}}$
+img: systems/rqg/assets/images/skills/boat.svg
+content: >
+  ${{openSystem.content.journalEntryType.skillDescription.title}}$
+  ${{openSystem.content.skill.boat.description}}$<hr><blockquote>${{openSystem.content.skillCategory.agility.description}}$</blockquote>
+flags:
+  rqg:
+    documentRqidFlags:
+      id: je..boat
+      lang: ${{openSystem.lang}}$
+      priority: ${{openSystem.prio}}$
+---
+# climb
+_id: C0qOlw6Iul5U3ZW3
+name: ${{openSystem.content.skill.climb.skillName}}$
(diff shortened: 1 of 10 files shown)


src/i18n/en/openSystem.json
```diff
@@ -1578,7 +1578,7 @@
...
+        "description": "<h3>Rune Spell</h3><p><i><b>Cost:</b> 3+<br><b>Type:</b> Ritual<br><b>Cults:</b> Issaries</i></p><p>Requires four wands. Creates a neutral trading ground 10 meters square, warded against vs. those with hostile intent. This provides the benefits of @JournalEntry[${{openSystem.module.name}}$.spirit-magic-descriptions-DUMMY]{Countermagic} 1,  @JournalEntry[${{openSystem.module.name}}$.spirit-magic-descriptions-DUMMY]{Spirit Screen} 2, and does 1D3 points of damage (ignoring armor) to a random hit location of an unwelcome intruder. Additional Rune points spent can increase the magical effects by the same amounts per point, or they can increase the size of the market by a multiple equal to those points spent. Lasts eight weeks.</p>"
```
wake42, line 1581: This kind of translated links does not work as it is now - and I think it will be hard to make the openSystem.json file translate itself so to speak. Let's leave this as it is, and I can think about how to manage this.

wake42 (changes requested):

src/assets/pack-templates/skill-descriptions/agility.yaml
```diff
@@ -0,0 +1,97 @@
+# boat
+_id: KRacu04VC87nhjOQ
```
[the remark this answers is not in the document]
TheChaosium: on it!

src/i18n/en/openSystem.json
TheChaosium: That has been there since before I started on the project. We do need to work how how add hyperlinks via RQIDs. We don't need to translate them, just link. Is that supported yet?

wake42: Not yet with RQID links - I plan to implement that in Foundry10 since they've changed stuff regarding that there.

wake42 (approved):

PR #362, opened 2022-09-13
rune descriptions now use labels

TheChaosium:


src/assets/pack-templates/rune-descriptions/rune-element-descriptions.yaml
```diff
@@ -1,171 +1,77 @@
...
-  <p><b>Color:</b> Black</p>
-  <p><b>Metal:</b> Lead</p>
-  <p><b>Phyla:</b> Insect</p>
+  ${{openSystem.content.journalEntryType.runeDescription.title}}$
+  <p><img style="float: right; margin-left:1rem; border:none;" src="systems/rqg/assets/images/runes/fire_sky.svg" height="175"></p>${{openSystem.content.runeDescription.fireskyElement.description}}$
 flags:
   rqg:
     documentRqidFlags:
-      id: je..darkness-element
+      id: je..firesky-element
```
wake42, line 49: the `/` in `fire/sky` should become `-`like this
```suggestion
      id: je..fire-sky-element
```

wake42 (changes requested):

src/assets/pack-templates/rune-descriptions/rune-element-descriptions.yaml
TheChaosium: ok. I will change in the rune and the rune description as the rune has firesky as well

wake42 (approved):

PR #370, opened 2022-09-14
add rune magic and descriptions

TheChaosium:
rune magic spells split out and added to the system

### DIFF
--- a/src/assets/pack-templates/rune-magic-spells/absorption.yaml
+++ b/src/assets/pack-templates/rune-magic-spells/absorption.yaml
@@ -18,6 +18,6 @@ data:
 flags:
   rqg:
     documentRqidFlags:
-      id: i.runeMagic.absorption
+      id: i.rune-magic.absorption
       lang: ${{openSystem.lang}}$
       priority: ${{openSystem.prio}}$
--- a/src/assets/pack-templates/rune-magic-spells/absorption.yaml
+++ b/src/assets/pack-templates/rune-magic-spells/absorption.yaml
@@ -1,0 +1,23 @@
+# absorption
+name: ${{openSystem.content.runeMagic.absorption.name}}$
+type: runeMagic
+img: systems/rqg/assets/images/items/rune-magic.svg
+data:
+  runes:
+    - Earth (element) # to deprecate
+    - Darkness (element) # to deprecate
+  points: 1
+  castingRange: ranged
+  duration: temporal
+  isStackable: true
+  isOneUse: false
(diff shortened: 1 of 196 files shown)


src/assets/pack-templates/rune-magic-spells/absorption.yaml
```diff
@@ -0,0 +1,23 @@
...
+  isStackable: true
+  isOneUse: false
+  chance: 0
+  descriptionRqidLink:
+    rqid: je..absorption
+    name: ${{openSystem.content.runeMagic.absorption.name}}$
+flags:
+  rqg:
+    documentRqidFlags:
+      id: i.runeMagic.absorption
```
wake42, line 21: rqid shouldn't be camel case. Goes for the other runemagic rqids as well.
```suggestion
      id: i.rune-magic.absorption
```

src/assets/pack-templates/rune-magic-spells/analyze-magic.yaml
```diff
@@ -0,0 +1,22 @@
...
+  isStackable: true
+  isOneUse: false
+  chance: 0
+  descriptionRqidLink:
+    rqid: je..analyze-magic
+    name: ${{openSystem.content.runeMagic.analyzeMagic.name}}$
+flags:
+  rqg:
+    documentRqidFlags:
+      id: i.runeMagic.rune-magic-analyze-magic
```
wake42, line 20: Double "rune magic" in id, applies to more rqids as well. 
```suggestion
      id: i.rune-magic.analyze-magic
```

wake42 (changes requested):

src/assets/pack-templates/rune-magic-spells/absorption.yaml
TheChaosium: Easy fix. I did't change it from the starter set version though so we missed it before. But for tomorrow.

src/assets/pack-templates/rune-magic-spells/analyze-magic.yaml
TheChaosium: Again, I just did file movement so this we missed before. I will review all spells and descriptions. The description links seems to work so it is most likely just the spells (and we don't link to them yet)

TheChaosium: Cleaned this up and removed links to other spells that will not work.

wake42 (approved):

PR #382, opened 2022-10-17
Using labels for hit location descriptions

TheChaosium:
Loaded with labels and also changed the routine that loads the hit locations
Note that we will need to change this per body type later

### DIFF
--- a/src/items/armor-item/armorSheet.ts
+++ b/src/items/armor-item/armorSheet.ts
@@ -5,9 +5,12 @@ import {
   armorTypeTranslationKeys,
   materialTranslationKeys,
 } from "../../data-model/item-data/armorData";
+import {
+  hitLocationTranslationKeys,
+} from "../../data-model/item-data/hitLocationData";
 import { equippedStatuses } from "../../data-model/item-data/IPhysicalItem";
 import { RqgItemSheet, RqgItemSheetData } from "../RqgItemSheet";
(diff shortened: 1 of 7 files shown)


src/items/armor-item/armorSheet.ts
```diff
@@ -49,7 +52,7 @@ export class ArmorSheet extends RqgItemSheet<ItemSheet.Options, ArmorSheetData |
       data: itemData,
       armorData: itemData.data,
       sheetSpecific: {
-        allHitLocations: getHitLocations(),
+        allHitLocations: hitLocationTranslationKeys.map((key) => localize(key)),
```
wake42, line 55: This is problematic since `getHitLocations()` reads the available hit locations from a system setting so that you can add more. There is a list of default (human) hitlocations here: https://github.com/sun-dragon-cult/fvtt-system-rqg/blob/master/src/system/settings/hitLocationNames.ts
And the starter set adds more hitlocations to that setting when you install it, see https://github.com/sun-dragon-cult/cha-rqg-fvtt-starterset/blob/master/src/scripts/moreHitlocations.mjs. Not sure how to make this translatable in a good way - I think this needs some redesign to make that possible.

TheChaosium: Yeah, that was a fear I had too... I think we need to load it based on supported body types (that can get added in different modules)

Let me think about it this weekend. I may be around a few nights

wake42:
This has been implemented in #440

PR #411, opened 2022-12-31
feat: allow players to add Passions

Moonpile:
Added a "+" button on the Passions tab of the Actor sheet that creates a new Passion on the Actor.  Also allow players to delete Passions via the context menu.

### DIFF
--- a/src/actors/rqgActorSheet.ts
+++ b/src/actors/rqgActorSheet.ts
@@ -1071,7 +1071,7 @@ export class RqgActorSheet extends ActorSheet<
           systemId,
           "defaultItemIconSettings"
         );
-        const newPassionName = localize("RQG.Actor.Passion.NewPassionName");
+        const newPassionName = localize("RQG.Item.Passion.PassionEnum.Loyalty");
         let passion = {
           name: newPassionName,
           type: "passion",
@@ -1079,7 +1079,7 @@ export class RqgActorSheet extends ActorSheet<
           system: { passion: newPassionName },
         };
         //@ts-ignore
-        await Item.createDocuments([passion], { parent: this.actor });
+        await Item.createDocuments([passion], { parent: this.actor, renderSheet: true });
       });
     });
   }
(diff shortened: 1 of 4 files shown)


src/actors/rqgActorSheet.ts
```diff
@@ -1062,6 +1062,26 @@ export class RqgActorSheet extends ActorSheet<
...
+        );
+        const newPassionName = localize("RQG.Actor.Passion.NewPassionName");
+        let passion = {
+          name: newPassionName,
+          type: "passion",
+          img: defaultItemIconSettings["passion"],
+          system: { passion: newPassionName },
+        };
+        //@ts-ignore
+        await Item.createDocuments([passion], { parent: this.actor });
```
wake42, line 1082: I think it would be good to open the itemsheet when the item is created. This can be done by adding `renderSheet: true` to the options.

src/actors/rqgActorSheet.ts
```diff
@@ -1062,6 +1062,26 @@ export class RqgActorSheet extends ActorSheet<
...
+    htmlElement?.querySelectorAll<HTMLElement>("[data-passion-add]").forEach((el) => {
+      el.addEventListener("click", async () => {
+        console.log("CLICK");
+        const defaultItemIconSettings: any = getGame().settings.get(
+          systemId,
+          "defaultItemIconSettings"
+        );
+        const newPassionName = localize("RQG.Actor.Passion.NewPassionName");
+        let passion = {
+          name: newPassionName,
```
wake42, line 1076: It would be nice to not need another translation just for this. Maybe one could pick one passion as a default and the player can then change it to what they want in the item sheet. Does "Loyalty" seems like a good default, feels like it's among the most common to add.

src/actors/rqgActorSheet.ts
Moonpile: Done

src/actors/rqgActorSheet.ts
Moonpile: Done

wake42 (approved):

PR #413, opened 2023-01-03
feat: hide "dead" actor health for spirits

Moonpile:
- actor.system.attributes.hitPoints can now be undefined
- hit points, if they exist, have a floor value of 3

### DIFF
--- a/src/actors/rqgActor.ts
+++ b/src/actors/rqgActor.ts
@@ -44,7 +44,7 @@ export class RqgActor extends Actor {
     const actorSystem = this.system;
     const { con, siz, pow } = this.actorCharacteristics();
     if (con !== null) {
-      if (actorSystem.attributes.hitPoints !== undefined) {
+      if (actorSystem.attributes.hitPoints != null) {
         actorSystem.attributes.hitPoints.max = RqgCalculations.hitPoints(con, siz, pow);
       } else {
         actorSystem.attributes.hitPoints = new Resource();
--- a/src/actors/rqgActor.ts
+++ b/src/actors/rqgActor.ts
@@ -13,6 +13,7 @@ import { systemId } from "../system/config";
 import { ResultEnum } from "../data-model/shared/ability";
 import { Rqid } from "../system/api/rqidApi";
 import { PrototypeTokenData } from "@league-of-foundry-developers/foundry-vtt-types/src/foundry/common/data/data.mjs";
+import { Resource } from "../data-model/shared/resource";
 
 export class RqgActor extends Actor {
   static init() {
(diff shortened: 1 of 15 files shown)


Moonpile:
Any Actor with no CON does not have Hit Points.  If an Actor has CON it has Hit Points even if it doesn't have SIZ or POW.  "actor.system.attributes.hitPoints" can now be undefined.  The health status of "dead" or "unconscious" will not be applied and therefore will not show up in the upper right of the Actor Sheet.

Note that applying the "Dead" or "Unconscious" effect to a token without changing its hit points will not show the health status in the upper right of the Actor Sheet.

Actors without CON show the "Health" section on the Combat tab with a note that the entity does not have hit points.

image

Actors with CON but without Hit Locations show the "Health" section with Total Hit points but with a note that the entity does not have Hit Locations.

image

Fixed an issue that arose from v10 transition where if an Actor had 1 or 2 Total Hit Points, and was therefore "unconscious", and you changed the hit points to zero (or lower), it was not automatically applying the "dead" status because of an error.  The same problem happened if "dead" and then you set the Total Hit Points to 1 or 2 and it would not set the status to "unconscious".

A few concerns that either can get addressed now if we have consensus or can be spun off into new issues:

- You can set Total Hit Points to negative values (not sure if appropriate or not)
- You can set Magic Points to negative values (definitely not appropriate)
- Setting Magic Points to zero does not set the health/token status to "unconscious" (assuming you're not also dead)

src/actors/rqgActor.ts
```diff
@@ -42,7 +43,17 @@ export class RqgActor extends Actor {
     super.prepareEmbeddedDocuments();
     const actorSystem = this.system;
     const { con, siz, pow } = this.actorCharacteristics();
-    actorSystem.attributes.hitPoints.max = RqgCalculations.hitPoints(con, siz, pow);
+    if (con !== null) {
```
wake42, line 46: At some places you check for undefined and other places for null. This can be unified with `value == null` (or `value != null`) since that will work for both `undefined` and `null`. Javascript apparently felt that null was such a great idea that they created two varieties of it :-)

src/actors/sheet-parts/health.hbs
```diff
@@ -1,43 +1,53 @@
...
+          <div class="ml-5 text-right" title="{{localize "RQG.Actor.Health.AddWound"}}" data-item-add-wound><i class="fas fa-hand-holding-medical"></i></div>
+        </div>
+      {{/each}}
+    </div>
+    {{else}}
+      <p>{{localize "RQG.Actor.Health.NoHitLocations"}}</p>
+    {{/if}}
+  {{else}}
+    <div class="flex-row nowrap flex-align-center">
+      <p>{{localize "RQG.Actor.Health.NoHitPoints"}}</p>
```
wake42, line 50: If an actor doesn't have hitpoints or hitlocations then I think it should just be blank. It's not an error and I wouldn't want that text on a spirit for example. If an actor don't have any health related things to show I think the health section should just not be there (including the health header).

wake42 (commented):
Adding undefined to these types with 
```ts
  value?: number;
  formula?: string;
```
https://github.com/sun-dragon-cult/fvtt-system-rqg/blob/1df823232cea355e55071264e0ae987134101585/src/data-model/actor-data/characteristics.ts#L2-L3 Vill generate a bunch of compile errors, but I think since these values actually can be undefined it would be good to define them as such and go through the places where it affects things and make active decisions. That is basically what you are doing in this PR but only for the effect of HP I think.

src/actors/sheet-parts/health.hbs
Moonpile: I didn't mean it to be an "error" so much as an explanation in case you were looking for it.  But I can change it.

Moonpile: Not that players are going be looking at spirits too much unless they have an associated spirit or fetch, but my players were definitely wondering where everything was when we started with just skills and hit locations.  I think they would have preferred to see the sections that were hidden.  So that was kind of my thinking.

src/actors/rqgActor.ts
Moonpile: Wow " !== null " does not pick up undefined but " != null" does.  Crazy.

Moonpile: If I want to set the attributes.hitPoints to "nothing" I should still do "attributes.hitPoints = undefined" though, right?

Moonpile: Done

src/actors/sheet-parts/health.hbs
Moonpile: Done

wake42:
Implemented in https://github.com/sun-dragon-cult/fvtt-system-rqg/pull/421

PR #416, opened 2023-01-07
feat: initialize Actor for new Token if Actor is unlinked

### LINKED ISSUE #415: Initialize Actor for new Token if Actor is unlinked

As a GM, when I drag an unlinked Actor onto a scene I want the system to automatically initialize all the characteristics of the temporary Actor attached to the new Token so that I can quickly create several similar but different Actors on a scene.  For example, dragging out five wolves and having slightly different characteristics for each one.

In Token.onCreate, if the Actor is not linked initialize the characteristics.

n/a

### Additional context

It might be wise to wrap this behavior in a setting flag for those who do not want it to behave this way.  May also consider settings flag for initializing characteristics when duplicating an unlinked actor.

Moonpile:
Fixes #415

### DIFF
--- a/src/combat/rqgToken.ts
+++ b/src/combat/rqgToken.ts
@@ -35,11 +35,7 @@ export class RqgToken extends Token {
     }
   }
 
-  //TODO: When upgrading @league-of-foundry-developers/foundry-vtt-types
-  //we can remove the ts-nocheck at the top of this file.
-  //In the meantime this allows us to get the userId which is now the third
-  //parameter.  We need it to ensure that we don't initialize the unlinked
-  //Actor for every logged in user.
+  // @ts-expect-error 
   protected _onCreate(
     options: PropertiesToSource<TokenDataProperties>,
     docModOptions: DocumentModificationOptions,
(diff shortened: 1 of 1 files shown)


Moonpile:
I can extend this to make the initialization behavior configurable in the RQG system settings, if you like.

src/combat/rqgToken.ts
```diff
@@ -28,4 +34,26 @@ export class RqgToken extends Token {
       });
     }
   }
+
+  //TODO: When upgrading @league-of-foundry-developers/foundry-vtt-types
+  //we can remove the ts-nocheck at the top of this file.
+  //In the meantime this allows us to get the userId which is now the third
+  //parameter.  We need it to ensure that we don't initialize the unlinked
+  //Actor for every logged in user.
```
wake42, line 42: I agree that it's problematic with the old Typescript types. My plan was to skip the intermediate versions and go directly to Foundry 10 version. Unfortunately the velocity of that project has dropped since the main developer has quit. There still is some movement so I'm still hoping for a usable version.

In the meantime I think it's better to use expect error like this and not have @ts-nocheck since those will just be forgotten about, while the @ts-expect-error will be an error when the ts types "fit again". 
```suggestion
  // @ts-expect-error 
```

That expect error is also needed on the CONFIG row.

wake42 (changes requested):

wake42 (approved):

Issue #74, opened 2021-10-15
Hide "dead" actor health for spirits


wake42:
Spirits currently get a "dead" actor health and a chat that soys something like `Spirit runs out of hitpoints and dies here and now!`

To fix that we need to type hitpoints.max as `number | undefined` and modify the code using it accordingly, setting actor health to "healthy" and handling the undefined value. 

Hide "health" panel if hitpoints.max is undefined - how should an added hitlocation be handled?

Moonpile:
I looked into this a bit and I don't think I fully understand TypeScript typing enough to tackle this.  However I had the following related observations and questions:
- If you have CON 1, SIZ 1, POW 1 the system currently gives you negative max hit points.  Since you need 3 hp to be conscious, is 3 hp the floor for hit points?
- If your current hit points are greater than your max hit points and you close and re-open the actor sheet, it forces your current hp back down to max hp. Vigor should allow you to have greater than max hp, though I think this will work by increasing CON.  Instead of recalculating(?), Vigor adds +3 to total hp and +1 to each location.

Moonpile:
We definitely need to think this through some more.  I think we need a way to say a creature does not have a Characteristic (null).  I suggest we make a GM context menu option to "turn off the characteristic".  Then, when calculating attributes such as hit points, if you have a null in any of the values the attribute depends upon, you would have null in the attribute.

Here's a Spirit I made with zero CON, SIZ, and POW.  I gave it hit locations because you can't see total hit points without hit locations.

![image](https://user-images.githubusercontent.com/3959760/210249928-7b74f1b8-47b9-46d8-8656-5fe8178c78ed.png)

And more realistically here's one with zero CON and SIZ but 15 POW.

![image](https://user-images.githubusercontent.com/3959760/210250395-fbcb5cc4-4360-4a88-8948-5df2b372dc8b.png)

Here's a Tiny thing with CON 1, SIZ 1, and POW 1 that I made, and as you can see it's got negative total hit points out of the box.  I think either the floor for hit points should be 3, or if your total is 3 or less that you're not unconscious or dead until you hit 0.    I'm not even sure what kind of creature we'd be modelling here though.  With stats like those it's basically not a physical threat of any kind.  Like a mouse would at least have more CON and maybe more POW.

![image](https://user-images.githubusercontent.com/3959760/210249994-ec582268-476d-453b-a23d-7516dac2b59c.png)

wake42:
[...]

PR #421, opened 2023-01-10
Moonpile+wake/hide dead actor health for spirits #74

wake42:
I started out with your branch @Moonpile but wanted to see what would happen if you define the characteristics as `number | unknown` and updated the system rqgCalculations to take the possibility of undefined characteristics into account. That, in combination with not removing `hitPoints` totally if the actor shouldn't have hitpoints but only accept that hitPoints.max is allowed to be undefined, made it possible to remove a lot of special case if's and simplify the logic.

### DIFF
--- a/src/system/rqgCalculations.ts
+++ b/src/system/rqgCalculations.ts
@@ -5,25 +5,38 @@ type LookupTableEntry<T> = {
 };
 
 export class RqgCalculations {
-  public static dexSR(dex: number): number | undefined {
+  public static dexSR(dex: number | undefined): number | undefined {
+    if (dex == null) {
+      return undefined;
+    }
     return this.lookup<number | undefined>(dex, dexSrTable);
   }
 
-  public static sizSR(siz: number): number | undefined {
+  public static sizSR(siz: number | undefined): number | undefined {
+    if (siz == null) {
+      return undefined;
+    }
     return this.lookup<number | undefined>(siz, sizSrTable);
   }
 
-  public static hitPoints(con: number, siz: number, pow: number): number {
-    let hp = con + RqgCalculations.linearMod(siz) / 5 + RqgCalculations.flattenedMod(pow) / 5;
-    if (hp < 3) {
+  public static hitPoints(
+    con: number | undefined,
+    siz: number | undefined,
+    pow: number | undefined
+  ): number | undefined {
+    let hp =
+      con != null && siz != null && pow != null
+        ? con + RqgCalculations.linearMod(siz) / 5 + RqgCalculations.flattenedMod(pow) / 5
+        : undefined;
+    if (hp && hp < 3) {
       // HP floor since 1 or 2 is unconscious.
       hp = 3;
     }
     return hp;
   }
 
-  public static damageBonus(str: number, siz: number): string {
-    if (str == null && siz == null) {
+  public static damageBonus(str: number | undefined, siz: number | undefined): string {
+    if (str == null || siz == null) {
       return "0";
     }
     const combined = str + siz;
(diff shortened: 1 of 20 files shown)


src/system/rqgCalculations.ts
```diff
@@ -5,20 +5,38 @@ type LookupTableEntry<T> = {
...
 
-  public static hitPoints(con: number, siz: number, pow: number): number {
-    return con + RqgCalculations.linearMod(siz) / 5 + RqgCalculations.flattenedMod(pow) / 5;
+  public static hitPoints(
+    con: number | undefined,
+    siz: number | undefined,
+    pow: number | undefined
+  ): number | undefined {
+    let hp =
+      con != null && siz != null && pow != null
```
wake42, line 28: Since you need con, siz & pow to calculate hitpoints I assume that you need all of them to have hitpoints. @Moonpile you only looked at con, but I think this makes more sense, do you agree?

Moonpile: I don't think you actually need SIZ and POW to have hit points.  They just modify the hit points as determined by CON.  If you don't have either SIZ or POW they just wouldn't be modified.  However, I'm not at all sure what someone without SIZ but **with** hit points would look like.  With SIZ and CON but no POW I would assume undead or a construct.

src/system/rqgCalculations.ts
```diff
@@ -5,20 +5,38 @@ type LookupTableEntry<T> = {
...
+      hp = 3;
+    }
+    return hp;
   }
 
-  public static damageBonus(str: number, siz: number): string {
-    if (str == null && siz == null) {
+  public static damageBonus(str: number | undefined, siz: number | undefined): string {
+    if (str == null || siz == null) {
       return "0";
```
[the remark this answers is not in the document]
Moonpile: I think this is getting into the question of what it means to not have either STR or SIZ.  There's a difference between 0 and not having it at all.   I think not having either STR or SIZ probably means you don't have a damage bonus because physical damage isn't a thing you can do anyway.  Having zero in either STR or SIZ or both means you have a physical form but are either incapacitated or permanently paralyzed or I suppose zero SIZ is just super tiny like a tardigrade or something.  I think there's a chart for the SIZ that might clear that up.

src/system/rqgCalculations.ts
Moonpile: Thinking about this again.  The real check happens in RqgActor line 46 because we want the entire hitPoints property to be undefined (or should it be null on line 55?) if you do not have CON.  If you have CON then you have hitPoints because they're based on CON and only modified by SIZ and POW.  

CON, SIZ, POW works (as the "normal" mode)
CON, no SIZ, POW works
CON, SIZ, no POW works
no CON = no hit points no matter the other two.

As is, this works properly to the best of my understanding of the rules.

src/system/rqgCalculations.ts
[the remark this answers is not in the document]
Moonpile: Thinking this through again with the code as it is right now:

STR 18, SIZ 18 = 1d6 Damage Bonus (makes sense)
no STR, SIZ 32 = 1d4 Damage Bonus (ok, I could buy this in that a SIZ 32 object could fall on you, but it should probably be "Does not inflict physical damage")
STR 32, no SIZ = 1d4 Damage Bonus (I really think this should be "Does not inflict physical damage".  It doesn't matter how strong you are if you don't have physical form)

I tried it with the change you're suggesting and it's the same exact result.

The ones with no STR or no SIZ should probably not be "0" (zero), but should be "does not inflict physical damage" or something like that.

PR #422, opened 2023-01-11
feat: evaluate Initialized Characteristics

### LINKED ISSUE #420: Evaluate Initialized Characteristics

**Note: I'm fully aware that this is a way out there feature request.**

As a GM, when I initialize all characteristics on an NPC, I'd like some quick visual indication about if this instance of the NPC has particularly good, normal, or particularly bad characteristics.  For example, the PCs are at a horse merchant and I drag several horses out into the stalls.  They're using Animal Lore to look for ones with high CON and STR and succeed so I can just quickly look at the visual indicators to pick out the "best" horse for them.

See above.

The alternative is doing some math in my head quickly or just changing the stats on the horse to be better.  However, if you have "4d6+12" it takes a moment to figure out what a "good" stat is versus a "very good" stat.

Moonpile:
Fixes #420

### DIFF
--- a/src/actors/_actorsheets.scss
+++ b/src/actors/_actorsheets.scss
@@ -266,44 +266,68 @@
   }
 
   input.characteristic-rank-high {
-    background-color: #018a31FF; //100%
-    color: palegoldenrod;
+    border-bottom: 7px solid #018a31FF; //100%
+    border-left: 5px solid transparent;
+    border-right: 5px solid transparent;
   }
   input.characteristic-rank-10 {
-    background-color: #018a31D6; //84%
+    border-bottom: 7px solid #018a31D6; //84%
+    border-left: 5px solid transparent;
+    border-right: 5px solid transparent;
   }
   input.characteristic-rank-9 {
-    background-color: #018a31AD; //68%
+    border-bottom: 7px solid #018a31AD; //68%
+    border-left: 5px solid transparent;
+    border-right: 5px solid transparent;
   }
   input.characteristic-rank-8 {
-    background-color: #018a3182; //51%
+    border-bottom: 7px solid #018a3182; //51%
+    border-left: 5px solid transparent;
+    border-right: 5px solid transparent;
   }
   input.characteristic-rank-7 {
-    background-color: #018a3159; //35%
+    border-bottom: 7px solid #018a3159; //35%
+    border-left: 5px solid transparent;
+    border-right: 5px solid transparent;
   }
   input.characteristic-rank-6 {
-    background-color: #018A3130; //19%
+    border-bottom: 7px solid #018A3130; //19%
+    border-left: 5px solid transparent;
+    border-right: 5px solid transparent;
   }
   input.characteristic-rank-5 {
-    background-color: #018A310D; //5%
+    border-bottom: 7px solid #018A310D; //5%
+    border-left: 5px solid transparent;
+    border-right: 5px solid transparent;
   }
   input.characteristic-rank-4 {
-    background-color: #8A01010D; //5%
+    border-bottom: 7px solid #8A01010D; //5%
+    border-left: 5px solid transparent;
+    border-right: 5px solid transparent;
   }
   input.characteristic-rank-3 {
-    background-color: #8A010130; //19%
+    border-bottom: 7px solid #8A010130; //19%
+    border-left: 5px solid transparent;
+    border-right: 5px solid transparent;
   }
   input.characteristic-rank-2 {
-    background-color: #8A010659; //35%
+    border-bottom: 7px solid #8A010659; //35%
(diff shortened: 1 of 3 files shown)


src/actors/_actorsheets.scss
```diff
@@ -264,4 +264,46 @@
...
+  input.characteristic-rank-1 {
+    background-color: #8A010182; //51%
+  }
+  input.characteristic-rank-0 {
+    background-color: #8A0101AD; //68%
+  }
+  input.characteristic-rank-low {
+    background-color: #8A0101FF;
+    color: palegoldenrod;
+  }
```
wake42, line 308: What do you think about something like this? Same colors, just not as a background.
```suggestion
  input.characteristic-rank-high {
    border-bottom: 7px solid #018a31FF; //100%
    border-left: 5px solid transparent;
    border-right: 5px solid transparent;
  }
  input.characteristic-rank-10 {
    border-bottom: 7px solid #018a31D6; //84%
    border-left: 5px solid transparent;
    border-right: 5px solid transparent;
  }
  input.characteristic-rank-9 {
    border-bottom: 7px solid #018a31AD; //68%
    border-left: 5px solid transparent;
    border-right: 5px solid transparent;
  }
  input.characteristic-rank-8 {
    border-bottom: 7px solid #018a3182; //51%
    border-left: 5px solid transparent;
    border-right: 5px solid transparent;
  }
  input.characteristic-rank-7 {
    border-bottom: 7px solid #018a3159; //35%
    border-left: 5px solid transparent;
    border-right: 5px solid transparent;
  }
  input.characteristic-rank-6 {
    border-bottom: 7px solid #018A3130; //19%
    border-left: 5px solid transparent;
    border-right: 5px solid transparent;
  }
  input.characteristic-rank-5 {
    border-bottom: 7px solid #018A310D; //5%
    border-left: 5px solid transparent;
    border-right: 5px solid transparent;
  }
  input.characteristic-rank-4 {
    border-bottom: 7px solid #8A01010D; //5%
    border-left: 5px solid transparent;
    border-right: 5px solid transparent;
  }
  input.characteristic-rank-3 {
    border-bottom: 7px solid #8A010130; //19%
    border-left: 5px solid transparent;
    border-right: 5px solid transparent;
  }
  input.characteristic-rank-2 {
    border-bottom: 7px solid #8A010659; //35%
    border-left: 5px solid transparent;
    border-right: 5px solid transparent;
  }
  input.characteristic-rank-1 {
    border-bottom: 7px solid #8A010182; //51%
    border-left: 5px solid transparent;
    border-right: 5px solid transparent;
  }
  input.characteristic-rank-0 {
    border-bottom: 7px solid #8A0101AD; //68%
    border-left: 5px solid transparent;
    border-right: 5px solid transparent;
  }
  input.characteristic-rank-low {
    border-bottom: 7px solid #8A0101FF;
    border-left: 5px solid transparent;
    border-right: 5px solid transparent;
  }
```

Moonpile:
@wake42 Yeah that looks more subtle and refined.

Moonpile:
The only thing is that I was using the goldenrod text color to make the "high" class easier to differentiate from the "10" class, and similarly "low" and "0".

image

So the 19 should be "high" and the 18 should be "10".  Hard to tell the diff.

Moonpile:
Some tests were failing, which was weird because it was building fine for me with no warnings.  The fixes made sense though.

Also I made the "high" class a bit darker and now, at least for me, it's distinguishable.

image

src/actors/rqgActorSheet.ts
```diff
@@ -240,7 +240,7 @@ export class RqgActorSheet extends ActorSheet<
       }
 
       // the tens value of the percentage of the value compared to the maxTotal
-      const rank = Math.floor(((char.value - minTotal) / (maxTotal - minTotal)) * 10);
+      const rank = Math.floor(((char?.value - minTotal) / (maxTotal - minTotal)) * 10);
```
wake42, line 243: Not at a computer, but what does `undefined - 3` evaluate as? The `char?.value` can be undefined and I'm thinking maybe doing `(char?.value ?? 0) ` is safer.

Moonpile: We would have done a "continue" in the for loop above if char or char.value is null, so it shouldn't be undefined at that point, right?

wake42: That's true so char should always have a value. Then there is no need to change it to `char?.value` right? Since that implies that char could be undefined.

Moonpile: Agreed.  char, char.value, minTotal, and maxTotal all must have values at that point, or else we would have hit a continue.

wake42 (approved):

PR #433, opened 2023-01-14
fix: null formula causes error

### LINKED ISSUE #432: bug: null formula causes error
### _What_: Describe the bug

Clearing (nulling) the formula for a characteristic causes an error.

### _How:_ To Reproduce

1. As GM, Open an Actor Sheet
2. Click in the characteristic formula and select the text.
3. Delete the text
4. Move the focus elsewhere
5. See error in console

### Expected behavior

No error should occur.  A characteristic without a formula cannot be evaluated, however, and should have no style associated with it.

### Screenshots

image

### Environment - please paste the information from foundry settings tab "support" button:

Current Master

### Additional context

Caused by #422

Moonpile:
Fixes #432

- Made sure characteristic has a numeric value before Improve Characteristic dialog
- Removed all code for putting roll cards into chat on any initialization of characteristics
- Ensure characteristic has valid formula before Initialize Characteristic and Initialize All Characteristics
- Ensure each characteristic has a valid formula before setting to Average


Moonpile:
At this point, none of the Characteristic Context Menu items throw an error regardless of whether the characteristic has a value or not, or a formula or not, or if the formula is a valid dice experession.

Moonpile:
Possible points to follow up on:

- If characteristic does not have a value, hide "Roll", "Quick Roll", and/or "Improve Characteristic"
- If characteristic does not have a valid dice expression in the formula, hide "Initialize Characteristic" (but not "Initialize All Characteristics")
- Provide some kind of message if the user enters an anything into a formula that is not empty string or a valid dice expression
- Now that we're not throwing dice cards into the chat there is no "feedback" from "Initialize Characteristic", "Initialize All Characteristics" and "Set All Characteristics to Average" except seeing the numbers change and they aren't necessarily going to change.  Perhaps just a Foundry Info?
- Can we include the Characteristic name so it's "Improve STR" and "Initialize STR", for example?

src/actors/context-menus/characteristic-context-menu.ts
```diff
@@ -84,7 +84,13 @@ export const characteristicMenuOptions = (
       const characteristic = actor.system.characteristics[charName];
       (characteristic as any).name = charName; // TODO adding extra properties that's not on type Characteristic
       const speakerName = token?.name ?? actor.prototypeToken.name ?? "";
-      showImproveCharacteristicDialog(actor, "characteristic", characteristic, speakerName);
+      if (
+        characteristic != null &&
+        characteristic.value != null &&
+        Number.isNumeric(characteristic.value)
+      ) {
```
wake42, line 91: I think the check for if there is a formula should be in the `condition` part so that you don't even see the context menu alternative if you don't have a value and formula.

Also I don't think it's necessary to check if characteristic is null since it should always be there.

This also applies to the `Initialize` & `Initialize All` context menus as well (except that you don't need a value for those menu items).

wake42 (changes requested):

wake42:
> * Now that we're not throwing dice cards into the chat there is no "feedback" from "Initialize Characteristic", "Initialize All Characteristics" and "Set All Characteristics to Average" except seeing the numbers change and they aren't necessarily going to change.  Perhaps just a Foundry Info?

I agree that some kind of feedback is good to have and a notification sounds like a good way of doing it. Would it be possible to just show one notification if you initialize all characteristics to avoid spamming.

wake42:
> * Can we include the Characteristic name so it's "Improve STR" and "Initialize STR", for example?

That would be a good improvement in UX - I'm all for that!

src/actors/context-menus/characteristic-context-menu.ts
Moonpile: I left this check here, but did it again in the condition for the "Improve Characteristic" context menu item.

I also did a similar check in the condition for "Initialize Characteristic" so that context menu item only displays if the characteristic has a valid formula.

Moonpile:
> I agree that some kind of feedback is good to have and a notification sounds like a good way of doing it. Would it be possible to just show one notification if you initialize all characteristics to avoid spamming.

Added ui notifications for "Initialize Characteristic", "Initialize All Characteristics", and "Set All Characteristics to Average".

Moonpile:
> > * Can we include the Characteristic name so it's "Improve STR" and "Initialize STR", for example?
> 
> That would be a good improvement in UX - I'm all for that!

Unfortunately, I don't think we can know the name of the characteristic in the "name" property of the Context Menu Item unless we override the ContextMenu object.  Is that possible or even worth doing?

wake42 (approved):
This solves the problem - lets leave any further improvements to the future.

PR #469, opened 2023-02-15
Initial German translation

phoenix-pnp:
Initial translation to german.
system.json @ 100%
openSystem.json @ 100%
Added martial arts to en as well, as it appears on the german standard char sheets.

### DIFF
--- a/src/combat/RqgCombatTracker.ts
+++ b/src/combat/RqgCombatTracker.ts
@@ -48,11 +48,15 @@ export class RqgCombatTracker extends CombatTracker {
       {
         name: localize("RQG.Foundry.CombatTracker.RemoveAllDuplicates"),
         icon: '<i class="fas fa-trash fa-fw"></i>',
-        callback: (li: JQuery) => {
+        callback: async (li: JQuery) => {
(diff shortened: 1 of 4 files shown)


src/i18n/de/openSystem.json
```diff
@@ -0,0 +1,2580 @@
...
+      "spirit": "Geist",
+      "stasis": "Stasis",
+      "summon": "Beschwören",
+      "tap": "Anzapfen",
+      "truth": "Wahrheit",
+      "water": "Wasser"
+    },
+    "runeDescription": {
+      "airElement": {
+        "name": "Luft (Element)",
```
wake42, line 1379: As you found out, the rune names unfortunately are not translatable currently. So to make the system work, they need to be the english names for now.

wake42 (changes requested):
I can't give any relevant review on the German translation as such, but I found a place where the translation key was translated as well.

I also found a bug in the code that hides compendium packs - the syntax seems to have changed since I wrote that code - I will include a fix for that in the system.

I'm trying to figure out if this can be included in the system - legal issues are not my strongest point. But for the time being I think it can live as a fork of the main system, until I've got some kind of confirmation that it can be included.

Issue #482, opened 2023-02-23
Minor Compendium Errors


Moonpile:
### Minor Compendium Errors

This is to keep track of minor compendium errors and typos.

Scent Intruder Journal Entry has the wrong rqid.
![image](https://user-images.githubusercontent.com/3959760/221026490-1a96373c-3ca6-401e-bf57-fb2e6eca7e9f.png)

Weapon skill Kick has the name "Melee Weapon"
![image](https://user-images.githubusercontent.com/3959760/221026781-caacfe66-8cd6-4195-b233-79ed587819f2.png)

wake42:
Hug is linked to the Kick skill
![image](https://user-images.githubusercontent.com/4853677/221044957-9a153632-e3c5-42eb-93f3-c1cca2dd34c0.png)

wake42:
AtlatlDart & AtlatlJavelin is missing a skill
![image](https://user-images.githubusercontent.com/4853677/221045303-14558e19-a811-40f5-93e8-e252693f59fe.png)
![image](https://user-images.githubusercontent.com/4853677/221045468-80e36bbd-227f-44bd-97f9-ed42a21c3602.png)

wake42:
Sling stones are equipped by default, should be carried
![image](https://user-images.githubusercontent.com/4853677/221046394-d7c1096e-7190-4b72-95fa-187b93c052ea.png)

PR #486, opened 2023-02-24
fix: Human-Template-include-items, Minor Compendium Errors, Multiple rqids with je..track

### LINKED ISSUE #481: Human Template include items

The human template does not include items properly.  Use the same item include functionality to include skills, runes, hit locations, etc.

Other related considerations:

1. Remove the "species-templates" compendium and move to "bestiary"?
2. Should there be a human template for each Homeland where we override the base chance of skills (RQG p 60-63) and add the "Homeland Base Passions" (RQG p 27), and anything else that changes between homelands?
3. Does it make sense to have "Player Character" templates that have some more things on them than NPC templates?

See above

n/a

### Additional context

n/a

Moonpile:
Fixes #481

- Used the Sartar text from the wiki in the Human (Sartarite)


Moonpile:
I created a vanilla "Human (Base)" and moved it to the bestiary.  This gets rid of the species-templates pack and breaks the Adventurer Wizard, but that is already well-broken.  I also created a "Human (Sartarite)" that has the cultural skills, rune, and passions pre-applied.  I didn't do the other homelands on the thinking that they will get included in the RQG module.

The "Customs ()" skills were using homelands.  I changed it to use cultures and agree with the skills as named in the "Homeland Cultural Skill Bonuses" from RQG p 60-63.  

Fixed the Kick skill name.

Removed the ui.notificication.errors in applyFnToItemFromHandlebarsArgs

Moonpile:
I accidentally fixed #482 on the same branch so I added it to this PR.

src/i18n/en/openSystem.json
```diff
@@ -2527,6 +2539,13 @@
       "bite": {
         "name": "Bite"
       },
+      "atlatl": {
```
wake42, line 2542: Should be plural
```suggestion
      "atlatls": {
```

wake42 (changes requested):
demo-actors & demo-items compendiums are still defined in system.json resulting in empty compendiums so they need to be removed.

Moonpile:
> demo-actors & demo-items compendiums are still defined in system.json resulting in empty compendiums so they need to be removed.

If we have a pack defined in system.json and there's not a corresponding directory in pack-templates, should we be getting a warning during build?  If so I think that can be a chore issue for another version.

src/i18n/en/openSystem.json
Moonpile: Fixed

wake42 (approved):

PR #494, opened 2023-02-25
fix: experience Checks should show anywhere a skill is displayed

### LINKED ISSUE #487: Experience Checks should show anywhere a skill is displayed
### _How:_ To Reproduce

1. Add a weapon.
2. Equip it
3. Give it's skill an experience check
4. Look at the combat tab and see the "checked" glow
5. Look at the Gear tab and you will not see the "checked" glow.

### Expected behavior

Anywhere a skill percentage is shown, it should glow if the skill is checked.

### Screenshots

Gear Tab:
image

Combat Tab:
image

### Environment - please paste the information from foundry settings tab "support" button:

2.1.0

### Additional context

n/a

Moonpile:
Fixes #487

- experience checks now show on the "Gear" tab
- can also roll from the "Gear" tab

### DIFF
--- a/src/actors/sheet-parts/gear-tab.hbs
+++ b/src/actors/sheet-parts/gear-tab.hbs
@@ -195,7 +195,11 @@
                         {{#if unusable}} warning{{/if}}"
                         {{#if unusable}}data-tooltip="{{localize "RQG.Actor.Combat.StrDexMinsNotMetTip"}}"{{/if}}
                   >
-                    {{skillchance @root.uuid skillId}}%&nbsp; {{skillname @root.uuid skillId}}
+                    <span class="combat contextmenu usage {{#if (eq @key "missile")}}missile{{/if}} {{experiencedclass @root.uuid skillId}}"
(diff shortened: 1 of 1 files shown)


wake42 (approved):
I'll approve, but I think we should change the full line from `Skill`, `Damage`, `STR`  to `DEX` be made clickable, since now the skill percentage is possible to use for attacking with the weapon. That could be fixed later - or added to this PR.

Moonpile:
> I'll approve, but I think we should change the full line from `Skill`, `Damage`, `STR` to `DEX` be made clickable, since now the skill percentage is possible to use for attacking with the weapon. That could be fixed later - or added to this PR.

While I agree with what you're saying, I think for now let's just go with what we have here for 2.1.1.  As I have said before, I think we might want to re-think the Combat/Combat grid and the Gear/Weapons grid.  I think we could boil it down to just one grid, and maybe put it somewhere where it can have the full width of the character sheet for more room for things like a drop down to quickly change ammo, and other things.  I think we need to put some thought into that though.

Issue #446, opened 2023-01-20
Easy Hit Location Roll


Moonpile:
### _Why:_ Is your feature request related to a problem?

As a GM or Player, I want to be able to quickly roll a hit location from the Actor sheet.

### _What:_ Describe the solution you'd like

A button or context menu in the Health and Hit Locations section that just immediately rolls a hit location on that actor and displays a result card in the chat.  It should have easy options for "high (1d10+10)", "low (1d10)", and "normal (1d20)".

### _Options:_ Describe alternatives you've considered

If you need a hit location and don't have a weapon card handy, this allows for a quick check.  For example, a dart trap fires a dart, where does it hit? 

### Additional context

n/a

PR #527, opened 2023-04-24
Moonpile/easy hit location roll #446

Moonpile:


Moonpile:
@wake42 What do we want to do with this?  Simplify the chat output so that it's just what kind of roll (low, normal, high) on whom, and what the result was?

PR #528, opened 2023-04-24
feat: add rune images in white

### LINKED ISSUE #519: Add Rune Images in White

As a GM, I want to be able to use Rune Images from "systems/rqg/assets/images/runes" as tiles on a scene or as tokens or other images, and I would like to have them available in a variety of colors.  If we provide the runes in white, I can use the tint feature to make them any color I want.

Render the runes in a variety of colors and have them in folders like "systems/rqg/assets/images/runes/white" or "systems/rqg/assets/images/runes/blue".  

And for Extra Credit, there's apparently some way to register a directory with Moulinette so that it's automatically hooked up.  Apparently Baileywiki modules do this with their images.  We might want to see if it's worth doing that as well.

This is entirely a "nice to have".

### Additional context

In Moulinette you can configure a source and point it to any folder full of images.  This is handy for the runes in particular but might be useful for other images we provide with the system.

image

Then you can view the images in that directory in Moulinette and drag them out as tiles or tokens.

image

Here I've placed a rune using a macro from Baileywiki to give it a drop shadow

image

And here's a white Air Rune with tint applied.

image

Moonpile:
Fixes #519

Added the runes in white.

Registered the runes directories with Moulinette if Moulinette is installed.


Moonpile:
I'm still looking into if there's a way to automatically register a directory with Moulinette that we don't mind including.

Moonpile:
https://github.com/SvenWerlen/moulinette-core/tree/main/docs

Moonpile:
> In the best of worlds I think the folder structure should be
> `assets/images/runes/black/...`
> `assets/images/runes/white/...`

I like this.  How bad would the migration script be?  I haven't really messed around with that.

> Also should all icons be available in white? then maybe they should be generated at compile time.

If you have Moulinette then you have access to all of the game-icons.net material automatically and you can download it in any color you want.  So there's not really any reason to include the icons we got from there except that we renamed a lot of them for our purposes.

I'm not a lawyer, but the license says we can adapt them so in my mind that covers renaming them and distributing them in that form.  https://game-icons.net/  https://creativecommons.org/licenses/by/3.0/

Not sure about the Chaosium IP on the runes.  However, the images I register that way are not sent to the cloud in any way.  They remain right where they were from installing the RQG system and the Moulinette mod simply adds them to its index.

wake42:
> > In the best of worlds I think the folder structure should be
> > `assets/images/runes/black/...`
> > `assets/images/runes/white/...`
> 
> I like this. How bad would the migration script be? I haven't really messed around with that.

This is an example of the previous move of the rune icons, you need to go through all items looking for rune images and change the `img` path to the new one: 
https://github.com/sun-dragon-cult/fvtt-system-rqg/blob/a2ab302f002dd317db4e457c41fe186482b03343/src/system/migrations/migrations-item/moveRuneIcon.ts#L4-L21 

We could maybe remove some of the older migrations by now, including that old `moveRuneIcons`.
https://github.com/sun-dragon-cult/fvtt-system-rqg/blob/a2ab302f002dd317db4e457c41fe186482b03343/src/system/migrations/migrateWorld.ts#L55-L68

wake42:
I accidentally wrote this in the issue instead of this PR so I'm copying it here...

Actually I'm starting to think that the default color for rune item images should be white, and the black one is a special case... If so then maybe we don't need this case. Instead I could modify the svgs so that I can color them black just in the actorSheet (using svg fill: currentColor;) to make it possible to style them to blend into the background (as they are now).

wake42:
Closed since the same #530 changes the default rqg icon color to white.

PR #554, opened 2023-05-31
feat: export Actor to Fillable PDF Character Sheet

### LINKED ISSUE #546: Export Actor to Fillable PDF Character Sheet

As a player or GM, I would like to be able to export actors to the Chaosium fillable character sheet.

A macro to facilitate this export.

We could bake this into the system api, but that makes it less adaptable by the end users.

### Additional context

n/a

Moonpile:
Fixes #546


Moonpile:
We could localize the text in the dialogs, but I figured that was a bit much.  Also if we were going to do that, we might want a different localization file for "macros included with the system" to just keep things organized.

PR #622, opened 2023-09-08
Moonpile/Unconscious-at-Zero-Magic-Points-#414

Moonpile:
When the actor is reduced to zero magic points (or less) their "displayed status" and token effect (if associated with a token) will be set to "unconscious".

Please note that if you open the character sheet from the sidebar instead of by double-clicking the token, the token effect will not be set.

This issue specified that "Magic Points should not be allowed to go into negative numbers", but I'm reconsidering the wisdom of that.  I think forcing negative values to zero will hide situations where the player is spending MP and the logic is not quite right.

### DIFF
--- a/src/actors/rqgActorSheet.ts
+++ b/src/actors/rqgActorSheet.ts
@@ -792,8 +792,12 @@ export class RqgActorSheet extends ActorSheet<
 
     // Hack: Temporarily change hp.value to what it will become so getCombinedActorHealth will work
     const hpTmp = this.actor.system.attributes.hitPoints.value;
+    const mpTmp = this.actor.system.attributes.magicPoints.value;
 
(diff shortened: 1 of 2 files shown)


wake42 (approved):
> This issue specified that "Magic Points should not be allowed to go into negative numbers", but I'm reconsidering the wisdom of that. I think forcing negative values to zero will hide situations where the player is spending MP and the logic is not quite right.

I read up on the rules and on page 244 it says:
image

contrast that to what it says about hit points on page 146:
image

It seems like there should be a cap on MP to not fall below 0. 

A player is prevented from casting magic for more MP than they have so they should never end up in negative MP because of that. In spirit combat you could get more MP damage than your current MP, but if I understand the rules correctly you still end up on 0.

That said this case, really does not make any asumptions about if MP can be negative or not so it's fine I think.

wake42 (approved):

PR #639, opened 2023-10-08
feat: add creationMode property to RqgActorData

Moonpile:


Moonpile:
Starting off really small!

Moonpile:
@wake42 So this was a very silly small PR.  I changed the property name to "editMode" based on your comments.

src/data-model/actor-data/rqgActorData.ts
```diff
@@ -14,6 +14,7 @@ export interface CharacterDataSourceData {
   // --- Derived / Convenience Data Below ---
   extendedName: string;
   attributes: Attributes; // Most are derived // TODO Split / move data?
+  editMode: boolean;
```
wake42, line 17: Just formatting, but both `editMode` and `extendedName` should be above the comment `// --- Derived / Convenience Data Below ---` since they are not derived

Moonpile: Done

wake42 (approved):

Issue #596, opened 2023-07-21
Migrate to use Foundry DataModel


wake42:
### _Why:_ Is your feature request related to a problem?

Change the definition of system data to use the [Foundry DataModel](https://foundryvtt.com/article/v10-data-model/) concept 

### Related
- [ ] #310

PR #667, opened 2023-11-25
feat: apply edit mode to sheet

Moonpile:


src/actors/rqgActorSheet.ts
```diff
@@ -1403,6 +1403,39 @@ export class RqgActorSheet extends ActorSheet<
         (createdItems[0] as RqgItem)?.sheet?.render(true);
       });
     });
+
```
Moonpile, line 1406: @wake42 these sections are the same or similar and could be generic if runes, skills, and passions all had the same interface or just the same properties.  I think this could be a good time to make a change for the future to support occupations and cultures:  (we could also finish this issue and do what I'm suggesting in another issue or as a part of the occupations and cultures changes)

.baseChance (for passions this would be 60, for skills it would vary)
.modifiers (maybe this is an array of key-value pairs?  Occupations and cultures and such will add permanent bonuses here)
.gainedChance (or maybe we could call it .experience or .experienceGain)
.categoryMod (only for skills, maybe even refactor this to just be another entry in modifiers)
.chance (the sum of the previous three)

src/actors/_actorsheets.scss
```diff
@@ -124,7 +124,7 @@
         .characteristics {
           grid-row: 2;
           grid-column: name-rune-start / name-rune-end;
-
+          width: 100%;
```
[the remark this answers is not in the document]
Moonpile: Done

src/actors/_actorsheets.scss
```diff
@@ -678,81 +678,94 @@
     }
   }
 
-  input.characteristic-rank-high {
+  *.characteristic-rank-high {
```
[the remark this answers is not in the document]
Moonpile: Done

Moonpile:
> In play mode I think we should hide the characteristics formula for everyone.

I can hide the dice expression formulas in play mode.  The green-red indicators are on that dice expression box.  As a gm I still want to see those in play mode so I know if the NPC is, say, large for its type, or weak for its type.


> Maybe we should show the "Edit Rune Magic Spell" context menu for trusted players.

I can do this, but should I also do it for Spirit Magic?  I'm not sure when even a trusted player is really going to need to edit either though, except for the spirit spell focus field which is always available.

> Should we only have one input field for runes & passions? Is it confusing with double percentages? Not sure... For skills it's good I think.

See my comment above about making runes, skills, and passions have similar fields so that we can reduce some handlers, but more importantly because I think it will be helpful to the later steps of the character creation process when you drag an occupation or tribe or culture onto your actor.  I think it would be best if it added the bonuses to the .modifiers property I'm proposing.  I also think that even for passions it make sense to have a .baseChance of 60, a .gainedChance so you know how much a player has gained if for no other reason.  On runes I want the .gainedChance because it will make it easier to determine if a player has correctly added their starting rune amounts (if we want to do validation of character creation).

src/actors/rqgActorSheet.ts
wake42: Not sure it makes sense to have a baseChance on runes & passions, especially not power runes that are matched against another rune.
The `data-rune-grid-edit` &. `data-passion-grid-edit` is the same thing though, I agree. For occupation / culture using this you mean there should be a modifier field for all these items that can be modified by those items, either by active effect or raw changing when you embed an occupation for example?
I guess we could implement a key-value structure and have a modifiers array, but arrays are not that easy to handle in Foundry dataModels I think. So if you wanted to link one of those to an input field it would take some extra code to accomplish.
Talking about datamodels that is also work that needs to be done - migrating the current definitions to use the datamodels Foundry uses: #596 
All in all I'm not sure it's good to squeeze all these items into the same mold - they are different I think (even if they are similar)

wake42:
> I can hide the dice expression formulas in play mode. The green-red indicators are on that dice expression box. As a gm I still want to see those in play mode so I know if the NPC is, say, large for its type, or weak for its type.

I was thinking that maybe we could have an individual setting for if you want to see the green-red indicator. That way each player / GM could decide for themselves if they want it. 

> > Maybe we should show the "Edit Rune Magic Spell" context menu for trusted players.
> 
> I can do this, but should I also do it for Spirit Magic? I'm not sure when even a trusted player is really going to need to edit either though, except for the spirit spell focus field which is always available.

You're right - lets leave it as it is now.

> 
> > Should we only have one input field for runes & passions? Is it confusing with double percentages? Not sure... For skills it's good I think.
> 
> See my comment above about making runes, skills, and passions have similar fields so that we can reduce some handlers, but more importantly because I think it will be helpful to the later steps of the character creation process when you drag an occupation or tribe or culture onto your actor. I think it would be best if it added the bonuses to the .modifiers property I'm proposing. I also think that even for passions it make sense to have a .baseChance of 60, a .gainedChance so you know how much a player has gained if for no other reason. On runes I want the .gainedChance because it will make it easier to determine if a player has correctly added their starting rune amounts (if we want to do validation of character creation).

Let's keep this as it is, and see where we land on this. It has a lot to do with character building as you say.

Moonpile:
> > I can hide the dice expression formulas in play mode. The green-red indicators are on that dice expression box. As a gm I still want to see those in play mode so I know if the NPC is, say, large for its type, or weak for its type.
> 
> I was thinking that maybe we could have an individual setting for if you want to see the green-red indicator. That way each player / GM could decide for themselves if they want it.

Ok, I'll do that tonight.  So would this be a new property on the actor like `.showCharacteristicRatings`?

Moonpile:
And generally I think we should see how this all works and we'll probably get feedback from users on "this is visually confusing" or "normal players should have access to X because of Y"

Moonpile:
@wake42 made a context menu for Show/Hide Characteristic Ratings and store the preference in actor.system.showCharacteristicRatings

So there was the thing with `margin-top` that you commented.  If that was intended to fix the difference in line height between displaying inputs for the characteristics and just showing the value in a div, it didn't seem to fix it.  I think we need to force the div that contains the input or div to always have a height of, say, 110px which is a bit taller than either, then center the contents (div or input).

wake42:
> So there was the thing with `margin-top` that you commented. If that was intended to fix the difference in line height between displaying inputs for the characteristics and just showing the value in a div, it didn't seem to fix it. I think we need to force the div that contains the input or div to always have a height of, say, 110px which is a bit taller than either, then center the contents (div or input).

Maybe the easiest way is to always show the input and add `pointer-events: none;` if it's not editable. Matching the height of a text to an input is hard, and the input will shrink if you make the sheet more narrow. The characteristics inputs doesn't look like an input anyway.

src/actors/rqgActorSheet.ts
```diff
@@ -223,6 +223,8 @@ export class RqgActorSheet extends ActorSheet<
       isEditable: this.isEditable,
       isGM: getGameUser().isGM,
       isPC: this.actor.hasPlayerOwner,
+      showCharacteristicRatings:
+        (getGame().settings.get(systemId, "showCharacteristicRatings") as boolean) || false,
```
wake42, line 227: You don't need to cast the type if you register the type of the setting in https://github.com/sun-dragon-cult/fvtt-system-rqg/blob/eabfc02b40cb75b1614b664a2e5d3e8466a1ae91/src/global.d.ts#L47

wake42 (commented):
I think the characteristics formula shouldn't be shown in play mode (for anyone, not even GM).

wake42 (approved):
Lets merge this and any details can be looked at later.

